PR #59, opened 2019-09-10
Alpha mainnet fix

jdkanani:


checkpoint/handler.go
```diff
@@ -151,20 +150,24 @@ func HandleMsgCheckpointAck(ctx sdk.Context, msg MsgCheckpointAck, k common.Keep
...
 			"endReceived", end,
 			"rootExpected", headerBlock.RootHash.String(),
 			"rootRecieved", root.String())
-
 		return common.ErrBadAck(k.Codespace).Result()
 	}
-
+	if headerBlock.EndBlock > end {
+		common.CheckpointLogger.Info("Adjusting endBlock to one already submitted on chain", "OldEndBlock", headerBlock.EndBlock, "AdjustedEndBlock", end)
+		headerBlock.EndBlock = end
```
mankenavenkatesh: Root of headerblock also need to be changed.

vaibhavchellani:
These concerns are correct. However this is an edge case, where `ACK` comes just before new proposer sends a `CHECKPOINT`. Nevertheless this has to be handled, this is like uncle block scenario.
- For rewards: We can divide rewards between both proposers on heidmall
- For fees: We can refund fees back to new proposer. 

The above mitigations could solve this.

vaibhavchellani:
This fix is going to stay on this branch. Closing PR.

PR #79, opened 2019-10-21
MAT 141 - Start bridge only when node is synced

vaibhavchellani:
Starts bridge services only when node is synced


vaibhavchellani:
@jdkanani @mankenavenkatesh please sqash and merge

PR #87, opened 2019-11-08
MAT-170 - Wiggle for heimdall tx

mankenavenkatesh:
Wiggle for heimdall tx. 
Approach- https://hackmd.io/055mjdcEQUa0uzAqMPLdNg

### DIFF
--- a/bridge/pier/common.go
+++ b/bridge/pier/common.go
@@ -20,8 +20,8 @@ import (
 	httpClient "github.com/tendermint/tendermint/rpc/client"
 	tmTypes "github.com/tendermint/tendermint/types"
 
-	"github.com/ethereum/go-ethereum/common"
 	"github.com/maticnetwork/heimdall/helper"
+	"github.com/maticnetwork/heimdall/types"
 	hmtypes "github.com/maticnetwork/heimdall/types"
 	rest "github.com/maticnetwork/heimdall/types/rest"
 )
@@ -89,7 +89,7 @@ func isProposer(cliCtx cliContext.CLIContext) bool {
 
 // IsStateSyncer returns if current user is state syncer or not
 func IsStateSyncer(cliCtx cliContext.CLIContext) bool {
-	var stateSyncerList []common.Address
+	var stateSyncerList []types.Validator
 	isStateSyncer := false
 
 	result, err := FetchFromAPI(cliCtx, GetHeimdallServerEndpoint(fmt.Sprintf(StateSyncerURL)))
@@ -104,8 +104,8 @@ func IsStateSyncer(cliCtx cliContext.CLIContext) bool {
 		return false
 	}
 
-	for _, syncerAddr := range stateSyncerList {
-		if bytes.Equal(syncerAddr.Bytes(), helper.GetAddress()) {
+	for _, syncer := range stateSyncerList {
+		if bytes.Equal(syncer.Signer.Bytes(), helper.GetAddress()) {
 			isStateSyncer = true
 			break
 		}
--- a/bridge/pier/common.go
+++ b/bridge/pier/common.go
@@ -20,6 +20,7 @@ import (
 	httpClient "github.com/tendermint/tendermint/rpc/client"
 	tmTypes "github.com/tendermint/tendermint/types"
 
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/maticnetwork/heimdall/helper"
 	hmtypes "github.com/maticnetwork/heimdall/types"
 	rest "github.com/maticnetwork/heimdall/types/rest"
@@ -39,6 +40,7 @@ const (
 	AccountDetailsURL     = "/auth/accounts/%v"
 	LastNoAckURL          = "/checkpoint/last-no-ack"
 	ProposersURL          = "/staking/proposer/%v"
+	StateSyncerURL        = "/clerk/state-syncer-list"
 	BufferedCheckpointURL = "/checkpoint/buffer"
 	LatestCheckpointURL   = "/checkpoint/latest-checkpoint"
 	CurrentProposerURL    = "/staking/current-proposer"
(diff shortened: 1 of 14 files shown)


bridge/pier/common.go
```diff
@@ -85,6 +87,32 @@ func isProposer(cliCtx cliContext.CLIContext) bool {
 	return false
 }
 
+// IsStateSyncer returns if current user is state syncer or not
+func IsStateSyncer(cliCtx cliContext.CLIContext) bool {
```
vaibhavchellani:
```suggestion
func IsStateSyncer(cliCtx cliContext.CLIContext) (isStateSyncer bool) {
```

bridge/pier/common.go
```diff
@@ -85,6 +87,32 @@ func isProposer(cliCtx cliContext.CLIContext) bool {
...
+	var stateSyncerList []types.Validator
+	isStateSyncer := false
```
vaibhavchellani: We can remove this, because we are anyway implicitly returning `false` and `true`

bridge/pier/common.go
```diff
@@ -85,6 +87,32 @@ func isProposer(cliCtx cliContext.CLIContext) bool {
...
+	err = json.Unmarshal(result.Result, &stateSyncerList)
+	if err != nil {
+		Logger.Error("Error unmarshalling state syncer", "error", err)
+		return false
+	}
+
+	for _, syncer := range stateSyncerList {
+		if bytes.Equal(syncer.Signer.Bytes(), helper.GetAddress()) {
+			isStateSyncer = true
+			break
```
vaibhavchellani: Directly return `true` here?

clerk/client/rest/query.go
```diff
@@ -66,3 +66,41 @@ func handlerRecordFn(
...
+
+func getNextStateSyncerHandlerFn(
+	cdc *codec.Codec,
+	cliCtx context.CLIContext,
+) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+
+		// fetch state syncer list
+		res, err := cliCtx.QueryWithData(fmt.Sprintf("custom/%s/%s/%s", clerkTypes.QuerierRoute, clerkTypes.QueryStateSyncer), nil)
+
```
vaibhavchellani: remove blank line

clerk/genesis.go
```diff
@@ -23,12 +23,16 @@ func DefaultGenesisState() GenesisState {
...
 	// add checkpoint headers
-	if len(data.EventRecords) != 0 {
+	eventCount := len(data.EventRecords)
+	if eventCount != 0 {
 		for _, record := range data.EventRecords {
 			keeper.SetEventRecord(ctx, *record)
 		}
 	}
+
+	keeper.SetStateSyncEventCount(ctx, uint64(eventCount))
```
vaibhavchellani: It might be the case where `keeper.SetEventRecord(ctx, *record)` will return an error but eventCount will not reflect that. Let's handle that error and/or ensure `EventCount` is always equal to actual events stored.

clerk/keeper.go
```diff
@@ -152,3 +152,43 @@ func (k *Keeper) IterateRecordsAndApplyFn(ctx sdk.Context, f func(record types.E
...
+	store.Set(StateSyncEventCountKey, eventCountInBytes)
+}
+
+// SetStateSyncEventCount sets state sync event count
+func (k *Keeper) SetStateSyncEventCount(ctx sdk.Context, eventCount uint64) {
+	store := ctx.KVStore(k.storeKey)
+
+	// convert state sync event to bytes
+	eventCountInBytes := []byte(strconv.FormatUint(eventCount, 10))
+
```
vaibhavchellani: remove line

clerk/querier_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+	type TestDataItem struct {
+		name               string
+		validatorCount     int
+		stateEventCount    uint64
+		expectedSyncerSize int
+		expectedValIDs     []types.ValidatorID
+	}
+
+	dataItems := []TestDataItem{
```
vaibhavchellani: Table tests!!! 🎉

clerk/genesis.go
mankenavenkatesh: Event Count is always equal to actual events stored.

clerk/querier.go
```diff
@@ -46,3 +50,40 @@ func handlerQueryRecord(ctx sdk.Context, req abci.RequestQuery, keeper Keeper) (
...
+	// if no of validators is less than three, return existing validators
+	if validatorCount < 3 {
+		syncerCount = validatorCount
+	}
+
+	// Select next 3 active validators
+	for i := 0; i < syncerCount; {
+		valIndex := eventCount % uint64(validatorCount)
+		validator, _ := stakingKeeper.GetValidatorFromValID(ctx, types.ValidatorID(valIndex+1))
+		if stakingKeeper.IsCurrentValidatorByAddress(ctx, validator.Signer.Bytes()) {
```
jdkanani: Can't we get validators from current validator set instead of getting all validators and filter by validator status?

bridge/pier/common.go
```diff
@@ -85,6 +87,32 @@ func isProposer(cliCtx cliContext.CLIContext) bool {
 	return false
 }
 
+// IsStateSyncer returns if current user is state syncer or not
+func IsStateSyncer(cliCtx cliContext.CLIContext) bool {
+	var stateSyncerList []types.Validator
+	isStateSyncer := false
+
+	result, err := FetchFromAPI(cliCtx, GetHeimdallServerEndpoint(fmt.Sprintf(StateSyncerURL)))
```
jdkanani: What if all three validators are offline?

clerk/querier.go
mankenavenkatesh: Yeah. we can do that. 
We should sort the current validator set by valID and do a binary search of validator Index. 
Sorting is extra time complexity. 

Considering CurrentValSet count is less, sorting should be fine i think.

bridge/pier/common.go
mankenavenkatesh: That would be the case if we go with backup model also.
Ideally, Instead of 3 validators We should increase the number to 1/3rd of currentvalidatorset count. It results in more number of transactions per sync.

Let me know if there is any better idea.

jdkanani: Yeah 1/3rd of current validator set would be better here and in bor as backup.

clerk/querier.go
jdkanani: @mankenavenkatesh  Validators in CurrentValSet will be already sorted as signer as far as I know. Can you check?

clerk/genesis.go
vaibhavchellani: oh yes, it has to be, but in this case in one case it might not be. The case where `SetStateSyncEventCount` throws an error.

vaibhavchellani:
@mankenavenkatesh Please resolve comments

PR #92, opened 2019-11-25
Added genesis checker

vaibhavchellani:


vaibhavchellani:
WIP: Testing underway.

cmd/heimdalld/main.go
```diff
@@ -223,6 +226,82 @@ func InitCmd(ctx *server.Context, cdc *codec.Codec) *cobra.Command {
...
+			}
+
+			// validate validators
+			validators := genesisState.StakingData.Validators
+			for _, v := range validators {
+				val, err := contractCaller.GetValidatorInfo(v.ID)
+				if err != nil {
+					return err
+				}
+				if val.VotingPower != v.VotingPower {
```
jdkanani: Please use the`GetPowerFromAmount()` method to get the amount before comparing to genesis state validator power. 

Check method: https://github.com/maticnetwork/heimdall/blob/38db2c8598ffce6594eb650565d089ff82703691/helper/util.go#L640-L647

jdkanani: I think you don't it since ValidatorInfo is doing before sending it from the contract.

staking/handler.go
```diff
@@ -91,7 +91,7 @@ func HandleMsgValidatorJoin(ctx sdk.Context, msg MsgValidatorJoin, k Keeper, con
 		ID:          validator.ID,
 		StartEpoch:  validator.StartEpoch,
 		EndEpoch:    validator.EndEpoch,
-		VotingPower: validator.VotingPower,
+		VotingPower: types.GetValidatorPower(validator.VotingPower),
```
jdkanani: No need to do that since validator already has voting power.

vaibhavchellani: Yes, I found 2 functions, was planning on removing one.

jdkanani:
Tested and fixed

PR #96, opened 2019-12-05
Cleanup

vaibhavchellani:
Removed the following
- Vesing Accounts
- Delegate Accounts
- Other functions which are not being used anwhere like Mint and Burn


supply/keeper.go
```diff
@@ -200,7 +200,7 @@ func (k Keeper) SendCoinsFromModuleToModule(
 }
 
 // SendCoinsFromAccountToModule transfers coins from an AccAddress to a ModuleAccount
-func (k Keeper) SendCoinsFromAccountToModule(
+func (k Keeper) c(
```
jdkanani: c? Typo?

PR #100, opened 2019-12-10
Dockerfile for CS1001

mankenavenkatesh:
This PR includes

Docker image for Counter Stage-0. Commit-id and config files are added in Dockerfile.

To run node, user just has to 
1. run the container.
2. update API_KEY, persistent_peers
3. start heimdall, rest-server


docker/Dockerfile.CS1001
```diff
@@ -0,0 +1,51 @@
+# Start from a Debian image with the latest version of Go installed
```
vaibhavchellani: Rename the file to something more generic.

Issue #142, opened 2020-01-12
Remove confirmation blocks check from msg handlers


vaibhavchellani:
Using these block confirmation checks might lead to nondeterministic results for nodes while might lead to consensus faults. 
Move these checks to `ante-handler` or `msg.ValidateBasic()`

PR #144, opened 2020-01-12
Check mainchain tx confirmation in ante

jdkanani:


jdkanani:
Fixes #142

auth/ante.go
```diff
@@ -55,12 +56,21 @@ type FeeCollector interface {
 	) sdk.Error
 }
 
+//
+// MainTxMsg tx hash
+//
+type MainTxMsg interface {
```
vaibhavchellani: Please use a better name for this struct.

vaibhavchellani: Suggestion: `BaseMsg`. Maybe we should make all our `msg` satisfy `BaseMsg` which would satisfy `sdk.Msg` as well.

jdkanani: `BaseMsg` shows the wrong terminology here. This tx should satisfy the Main chain tx.

PR #145, opened 2020-01-12
Fix #143: Add rest endpoints to display different lists

### LINKED ISSUE #143: Add rest endpoints to display different lists
We need to have an API to various types of lists. 

For example:
- [ ] All spans
- [ ] All records synced by `clerk`
- [x] All headers

jdkanani:
Fix #143


clerk/client/rest/query.go
```diff
@@ -56,3 +61,48 @@ func handlerRecordFn(cliCtx context.CLIContext) http.HandlerFunc {
...
+		}
+
+		// get page
+		page, ok := rest.ParseUint64OrReturnBadRequest(w, vars.Get("page"))
+		if !ok {
+			return
+		}
+
+		// get limit
+		limit, ok := rest.ParseUint64OrReturnBadRequest(w, vars.Get("limit"))
```
vaibhavchellani: So the limit is user defined?

clerk/keeper.go
```diff
@@ -119,6 +119,32 @@ func (k *Keeper) GetAllEventRecords(ctx sdk.Context) (records []*types.EventReco
 	return
 }
 
+// GetEventRecordList returns all records with params like page and limit
+func (k *Keeper) GetEventRecordList(ctx sdk.Context, page uint64, limit uint64) ([]types.EventRecord, error) {
```
vaibhavchellani: Would be better I think if the keeper didn't need page and limit params. Can we add this logic in the querier instead of keeper?

clerk/types/querier.go
```diff
@@ -14,3 +15,14 @@ type QueryRecordParams struct {
 func NewQueryRecordParams(recordID uint64) QueryRecordParams {
 	return QueryRecordParams{RecordID: recordID}
 }
+
+// QueryRecordListParams defines the params for querying accounts.
```
vaibhavchellani: How about we abstract this out and use the same struct for `checkpoint/list`, `span/list` and `clerk/list`

vaibhavchellani:
@jdkanani please check out the comments.

clerk/client/rest/query.go
jdkanani: Yes. It has also max limit.

clerk/keeper.go
jdkanani: Storage should read only needed data. We shouldn't fetch all data from storage and filter at querier layer.

clerk/types/querier.go
jdkanani: Fixed.

clerk/keeper.go
vaibhavchellani: Agreed.

PR #157, opened 2020-01-21
MAT-533 - don't propose checkpoint if current validator's checkpoint …

mankenavenkatesh:
…is present in buffer

checkpoint optimization - 
currently validator keeps proposing checkpoint transactions with increased checkpoint size on each poll interval increasing the fee cost. 
validator should propose new checkpoint only if his checkpoint is not in buffer.

### DIFF
--- a/bridge/pier/checkpoint.go
+++ b/bridge/pier/checkpoint.go
@@ -282,7 +282,7 @@ func (c *Checkpointer) sendRequest(newHeader *types.Header) {
 	end := expectedCheckpointState.newEnd
 
 	if bufferedCheckpoint != nil && bufferedCheckpoint.proposer.Equals(hmtypes.BytesToHeimdallAddress(helper.GetAddress())) && bufferedCheckpoint.start == start {
-		c.Logger.Info("Checkpoint already proposed by current validator with same start block", "validatorAddress", bufferedCheckpoint.proposer, "startBlock", start)
+		c.Logger.Info("Checkpoint already proposed to heimdall by us", "proposer", bufferedCheckpoint.proposer, "startBlock", start, "endBlock", end)
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


jdkanani:
Shouldn't it override if the current buffer is expired?

bridge/pier/checkpoint.go
```diff
@@ -280,6 +280,12 @@ func (c *Checkpointer) sendRequest(newHeader *types.Header) {
 
 	start := expectedCheckpointState.newStart
 	end := expectedCheckpointState.newEnd
+
+	if bufferedCheckpoint != nil && bufferedCheckpoint.proposer.Equals(hmtypes.BytesToHeimdallAddress(helper.GetAddress())) && bufferedCheckpoint.start == start {
+		c.Logger.Info("Checkpoint already proposed by current validator with same start block", "validatorAddress", bufferedCheckpoint.proposer, "startBlock", start)
```
vaibhavchellani: Might be a good idea to trim this log and more checkpoint info
```suggestion
		c.Logger.Info("Checkpoint already proposed to heimdall by us", "proposer", bufferedCheckpoint.proposer, "startBlock", start,"endBlock",end)
```

vaibhavchellani:
@jdkanani I think the idea is if a proposer has already proposed a checkpoint there should be no need to propose on heimdall again. 
However it might be a good idea to replace the checkpoint in buffer with a bigger checkpoint post `NoAckTimeInterval * 3`, but again that's just an optimisation on the bridge side, that won't change anything, will just ensure that we try again with a bigger checkpoint.

vaibhavchellani:
I would suggest keeping this PR minimal and moving ahead at this point of time.

jdkanani:
@vaibhavchellani should we merge this?

Issue #208, opened 2020-02-25
Validator signer change makes older validator-join transactions invalid


jdkanani:
Since both `MsgValidatorJoin` transactions are using the latest contract state for validator information, signer change on staking contracts makes older validator-join invalid.

This results in sync failure.

Similar to #206

jdkanani:
Fixed in `develop` branch.

PR #209, opened 2020-02-25
Fixes #207 and #208

### LINKED ISSUE #207: Log index is invalid from validator-join cli command

jdkanani:
Fix invalid log index while sending validator join tx and use event log instead of contract state for validator join

### DIFF
--- a/helper/call.go
+++ b/helper/call.go
@@ -37,6 +37,7 @@ type IContractCaller interface {
 	GetConfirmedTxReceipt(common.Hash) (*ethTypes.Receipt, error)
 	GetBlockNumberFromTxHash(common.Hash) (*big.Int, error)
 	DecodeValidatorTopupFeesEvent(*ethTypes.Receipt, uint64) (*stakinginfo.StakinginfoTopUpFee, error)
+	DecodeValidatorJoinEvent(*ethTypes.Receipt, uint64) (*stakinginfo.StakinginfoStaked, error)
 	DecodeValidatorStakeUpdateEvent(*ethTypes.Receipt, uint64) (*stakinginfo.StakinginfoStakeUpdate, error)
 	DecodeNewHeaderBlockEvent(*ethTypes.Receipt, uint64) (*rootchain.RootchainNewHeaderBlock, error)
 	DecodeSignerUpdateEvent(*ethTypes.Receipt, uint64) (*stakinginfo.StakinginfoSignerChange, error)
@@ -46,6 +47,7 @@ type IContractCaller interface {
 	StakeFor(common.Address, *big.Int, *big.Int, bool) error
 
 	CurrentAccountStateRoot() ([32]byte, error)
+	FindStakedEventLogIndex(*ethTypes.Receipt) (uint64, bool)
 
 	// bor related contracts
 	CurrentSpanNumber() (Number *big.Int)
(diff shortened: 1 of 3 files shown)


helper/call.go
```diff
@@ -361,6 +363,28 @@ func (c *ContractCaller) DecodeValidatorStakeUpdateEvent(receipt *ethTypes.Recei
 	return event, nil
 }
 
+// DecodeValidatorJoinEvent represents validator staked event
+func (c *ContractCaller) DecodeValidatorJoinEvent(receipt *ethTypes.Receipt, logIndex uint64) (*stakinginfo.StakinginfoStaked, error) {
+	event := new(stakinginfo.StakinginfoStaked)
+
+	found := false
+	for _, vLog := range receipt.Logs {
+		if uint64(vLog.Index) == logIndex {
```
vaibhavchellani: The log index would be fixed for ValidatorJoin log, right?

vaibhavchellani: If yes, we can remove one parameter.

helper/call.go
```diff
@@ -474,6 +498,25 @@ func (c *ContractCaller) EncodeStateSyncedEvent(log *ethTypes.Log) (*statesender
 	return event, nil
 }
 
+// FindStakedEventLogIndex find log index for staked event
+func (c *ContractCaller) FindStakedEventLogIndex(receipt *ethTypes.Receipt) (uint64, bool) {
+	abiObject := &c.StakingInfoABI
+	eventName := "Staked"
```
vaibhavchellani: Probably add this string as a const and use in both `FindStakedEventLogIndex` and `DecodeValidatorJoinEvent`, I think we have a list of constants in some package.

staking/handler.go
```diff
@@ -34,42 +34,37 @@ func NewHandler(k Keeper, contractCaller helper.IContractCaller) sdk.Handler {
...
-	// fetch validator from mainchain
-	validator, err := contractCaller.GetValidatorInfo(msg.ID)
-	if err != nil {
-		k.Logger(ctx).Error(
-			"Unable to fetch validator from rootchain",
-			"error", err,
-		)
-		return hmCommon.ErrNoValidator(k.Codespace()).Result()
+	// TODO remove this block in next release and use msg.LogIndex -- start
+	logIndex, found := contractCaller.FindStakedEventLogIndex(receipt)
```
vaibhavchellani: Can we have a parent function which does both `FindAndDecodeEventLog`, just to remove clutter from handler.

vaibhavchellani:
I think at this point it's prob best to review all the other areas we are using contracts without any specific anchors like `logs` or `blockheights` and then remove functions like `GetValidatorInfo` completely.

helper/call.go
jdkanani: No. `LogIndex` is against the block not the tx. The same misconception introduced this bug in `ValidatorJoin` command.

helper/call.go
jdkanani: We will remove `FindStakedEventLogIndex` method in cs-2003 and only use `DecodeValidatorJoinEvent` using `msg.LogIndex`.

staking/handler.go
jdkanani: This will get removed in cs-2003 and only `DecodeValidatorJoinEvent` will be used. We cannot remove `DecodeValidatorJoinEvent` due to different struct types.

jdkanani:
@vaibhavchellani Yes. Plan is to remove `contractCaller.GetValidatorInfo` from Topup as well. But, sometimes Topup tx won't have validator in the state, `signer` address is needed in `TopupEvent`. 

cc @0xAshish

jdkanani:
Referencing #207 & #208 for more details

helper/call.go
vaibhavchellani: yep: https://github.com/ethereum/go-ethereum/blob/master/core/types/log.go#L50

PR #218, opened 2020-02-26
Fix: Remove contract state call from topup

jdkanani:
- This PR removes a contract state call from Topup handler. Now it depends on the event log from the Ethereum chain and updates Heimdall state.
- `TopUpEvent` on the Ethereum contract has been updated to include `signer`. This PR updates ABI and decodes events based on that.
    There are two types of Topup (from overview):
    1.  Before validator joins - in this case, handler mints fee tokens to `signer` from the `eventLog.Signer`
    2. After validator joins (if validator info is Heimdall state) - handler mints fee tokens to `state.Validator.Signer`
    By doing that, we are preventing race condition when multiple events, `TopUpFee` and `SignerChange` from Ethereum occur. Reason: If we always mint fee tokens to `eventLog.Signer`. And, if `TopUp` and `SignerChange` both occurred on the Ethereum chain in that order. But, the bridge would process `SignerChange` first and `TopUp` after that, then the old signer will get the fee-tokens. 
- Fix overflow issue #221 by always using `Big.Int` for stake amount


jdkanani:
Fixes #221

app/app.go
```diff
@@ -124,6 +125,31 @@ func (d CrossCommunicator) IsCurrentValidatorByAddress(ctx sdk.Context, address
 	return d.App.StakingKeeper.IsCurrentValidatorByAddress(ctx, address)
 }
 
+// AddFeeToDividendAccount add fee to dividend account
+func (d CrossCommunicator) AddFeeToDividendAccount(ctx sdk.Context, valID types.ValidatorID, fee *big.Int) sdk.Error {
```
vaibhavchellani: Can we change`d` to something else?

staking/keeper.go
```diff
@@ -27,8 +28,12 @@ var (
 	StakingSequenceKey        = []byte{0x24} // prefix for each key for staking sequence map
 )
 
-type AckRetriever interface {
+// ModuleCommunicator manages different module interaction
+type ModuleCommunicator interface {
```
vaibhavchellani: Please elaborate on the purpose of this module, how do we decide what goes in here etc

jdkanani:
@vaibhavchellani fixed. Also opened #226 cc @ptsayli

PR #231, opened 2020-03-03
Refactor topup module

ptsayli:
- [x] Create new topup module, removed from the bank module.
- [x] Add cli commands for query and tx top-ups
- [x] Allow partial fee withdrawals


topup/client/rest/tx.go
```diff
@@ -0,0 +1,97 @@
...
+
+	bankTypes "github.com/maticnetwork/heimdall/bank/types"
+	restClient "github.com/maticnetwork/heimdall/client/rest"
+	"github.com/maticnetwork/heimdall/types"
+	"github.com/maticnetwork/heimdall/types/rest"
+)
+
+// RegisterRoutes - Central function to define routes that get registered by the main application
+func RegisterRoutes(cliCtx context.CLIContext, r *mux.Router) {
+	r.HandleFunc("/bank/accounts/topup", TopupHandlerFn(cliCtx)).Methods("POST")
```
vaibhavchellani: Change the API's

topup/genesis.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/maticnetwork/heimdall/topup/types"
+)
+
+// InitGenesis sets distribution information for genesis.
+func InitGenesis(ctx sdk.Context, keeper Keeper, data types.GenesisState) {
+	// TODO: change later
+	// if data.TopupSequence > 0 {
```
vaibhavchellani: Is this how it's supposed to be?

topup/handler.go
```diff
@@ -0,0 +1,161 @@
...
+	if k.HasTopupSequence(ctx, sequence) {
+		k.Logger(ctx).Error("Older invalid tx found")
+		return hmCommon.ErrOldTx(k.Codespace()).Result()
+	}
+
+	// add total topups amount
+	topupObject.TotalTopups = topupObject.TotalTopups.Add(topupAmount)
+
+	// increase coins in account
+	if _, ec := k.bk.AddCoins(ctx, signer, topupAmount); ec != nil {
```
vaibhavchellani: Please use more descriptive name instead of `ec`

topup/keeper.go
```diff
@@ -0,0 +1,153 @@
...
+	// return true if validator
+	return &validatorTopup, nil
+}
+
+// SetValidatorTopup sets validator topup object
+func (keeper Keeper) SetValidatorTopup(ctx sdk.Context, addr hmTypes.HeimdallAddress, validatorTopup types.ValidatorTopup) error {
+	store := ctx.KVStore(keeper.key)
+
+	// validator topup
+	bz, err := types.MarshallValidatorTopup(keeper.cdc, validatorTopup)
```
vaibhavchellani: Nitpicky comment, feel free to ignore but this method has no benefit, same amount of code is written in UnmarshallValidatorTopup, it's better we marshall inline, we will save on 2 variables.

topup/module.go
```diff
@@ -0,0 +1,159 @@
...
+
+	"github.com/maticnetwork/heimdall/topup/types"
+	hmTypes "github.com/maticnetwork/heimdall/types"
+)
+
+var (
+	_ module.AppModule            = AppModule{}
+	_ module.AppModuleBasic       = AppModuleBasic{}
+	_ hmTypes.HeimdallModuleBasic = AppModule{}
+	// _ module.AppModuleSimulation = AppModule{}
```
vaibhavchellani: Remove comment

topup/types/msg.go
```diff
@@ -0,0 +1,126 @@
...
+		TxHash:      txhash,
+		LogIndex:    logIndex,
+	}
+}
+
+// Route Implements Msg.
+func (msg MsgTopup) Route() string { return RouterKey }
+
+// Type Implements Msg.
+func (msg MsgTopup) Type() string { return "topup" }
```
vaibhavchellani: Use `ModuleName` from `types/keys.go`?

topup/keeper.go
jdkanani: @ptsayli as we discussed please remove validator top-up from the store. We are not using it.

topup/handler.go
ptsayli: umm, not sure
Can you please suggest? :grin:

topup/keeper.go
ptsayli: Removed `ValidatorTopup`

topup/handler.go
jdkanani: `err`?

topup/types/genesis.go
```diff
@@ -2,19 +2,21 @@ package types
 
 // GenesisState is the bank state that must be provided at genesis.
 type GenesisState struct {
-	// TopupSequences map[uint256]bool `json:"topup_sequence" yaml:"topup_sequence"`
+	TopupSequences []*uint64 `json:"tx_sequences" yaml:"tx_sequences"`
```
jdkanani: If you are using a pointer, you need to check if the values in the array are not nil while importing genesis. There is a method called `ValidateGenesis`. Please use that for validation.

topup/handler.go
vaibhavchellani: yep

topup/keeper.go
```diff
@@ -0,0 +1,129 @@
...
+
+var (
+	// DefaultValue default value
+	DefaultValue = []byte{0x01}
+	// TopupSequencePrefixKey represents topup sequence prefix key
+	TopupSequencePrefixKey = []byte{0x81}
+)
+
+// ModuleCommunicator manager to access validator info
+type ModuleCommunicator interface {
```
jdkanani: You don't need a module communicator since you have a staking keeper in the topup keeper.

PR #235, opened 2020-03-08
Add params and gov module

jdkanani:
**Changes**

Modules

- [x] Adds `params` module. Now, `params` in each module can be changed using `gov` module's `submit-proposal`
- [x] Adds `gov` module. The governance module uses validator id as the primary identifier for proposal, deposit, and vote.

Decoding/Encoding

- [x] Uses RLP to decode/encode only `MsgCheckpoint` message. For other messages, use default amino encoding.

**Fixes**

Genesis

- [x] Fixes module account export/import


app/app.go
```diff
@@ -180,31 +183,31 @@ func NewHeimdallApp(logger log.Logger, db dbm.DB, baseAppOptions ...func(*bam.Ba
...
 
 	// init params keeper and subspaces
-	app.ParamsKeeper = params.NewKeeper(app.cdc, keys[params.StoreKey], tkeys[params.TStoreKey], params.DefaultCodespace)
+	app.ParamsKeeper = params.NewKeeper(app.cdc, keys[paramsTypes.StoreKey], tkeys[paramsTypes.TStoreKey], paramsTypes.DefaultCodespace)
 	app.subspaces[authTypes.ModuleName] = app.ParamsKeeper.Subspace(authTypes.DefaultParamspace)
 	app.subspaces[bankTypes.ModuleName] = app.ParamsKeeper.Subspace(bankTypes.DefaultParamspace)
 	app.subspaces[supplyTypes.ModuleName] = app.ParamsKeeper.Subspace(supplyTypes.DefaultParamspace)
-	// app.subspaces[gov.ModuleName] = app.ParamsKeeper.Subspace(gov.DefaultParamspace).WithKeyTable(gov.ParamKeyTable())
+	app.subspaces[govTypes.ModuleName] = app.ParamsKeeper.Subspace(govTypes.DefaultParamspace).WithKeyTable(govTypes.ParamKeyTable())
 	app.subspaces[stakingTypes.ModuleName] = app.ParamsKeeper.Subspace(stakingTypes.DefaultParamspace)
```
mankenavenkatesh: later when we are creating staking keeper at line no 248, we are passing **app.subspaces[stakingTypes.ModuleName]** as argument. 

Inside keeper constructor, we are setting **paramSpace:  paramSpace.WithKeyTable(types.ParamKeyTable())**
WithKeyTable function returns a new subspace with modified table adding staking params. 
so there will be mismatch of staking subspace in paramskeeper and stakingkeeper 

so we should store pointer to the subspace i think in the map in app.go.

mankenavenkatesh: Also do we need to have subspace for module(eg...staking) both in stakingkeeper and paramskeeper??

gov/proposal.go
```diff
@@ -0,0 +1,164 @@
...
+	if !keeper.router.HasRoute(content.ProposalRoute()) {
+		return types.Proposal{}, types.ErrNoProposalHandlerExists(keeper.codespace, content)
+	}
+
+	// Execute the proposal content in a cache-wrapped context to validate the
+	// actual parameter changes before the proposal proceeds through the
+	// governance process. State is not persisted.
+	cacheCtx, _ := ctx.CacheContext()
+	handler := keeper.router.GetRoute(content.ProposalRoute())
+	if err := handler(cacheCtx, content); err != nil {
```
mankenavenkatesh: why are we calling handler here on submit proposal? 

In my understand, handler should be called only in end-blocker once required votes are received. handler will update the params.

jdkanani: It validates if the handler will actually execute when the proposal succeeds. The state change is not be persisted since it is cache context.

FYI: This file is borrowed from cosmos-sdk `x/gov` package.

mankenavenkatesh: Got it. Thanks.

app/app.go
jdkanani: That is modifying the same subspace with `WithKeyTable`

gov/types/codec.go
```diff
@@ -0,0 +1,42 @@
...
+// RegisterCodec registers all the necessary types and interfaces for
+// governance.
+func RegisterCodec(cdc *codec.Codec) {
+	cdc.RegisterInterface((*Content)(nil), nil)
+
+	cdc.RegisterConcrete(MsgSubmitProposal{}, "gov/MsgSubmitProposal", nil)
+	cdc.RegisterConcrete(MsgDeposit{}, "gov/MsgDeposit", nil)
+	cdc.RegisterConcrete(MsgVote{}, "gov/MsgVote", nil)
+
+	cdc.RegisterConcrete(TextProposal{}, "heimdall/TextProposal", nil)
```
mankenavenkatesh: we are handling only ParameterChangeProposal

Please remove (SoftwareUpgradeProposal, TextProposal)

bor/keeper.go
```diff
@@ -216,7 +217,8 @@ func (k *Keeper) SelectNextProducers(ctx sdk.Context) (vals []hmTypes.Validator,
 
 	// spanEligibleVals are current validators who are not getting deactivated in between next span
 	spanEligibleVals := k.sk.GetSpanEligibleValidators(ctx)
-	producerCount, err := k.GetProducerCount(ctx)
+	producerCount := k.GetParams(ctx).ProducerCount
+	fmt.Println("producerCount", producerCount)
```
mankenavenkatesh: pls remove this

bridge/pier/span.go
```diff
@@ -129,6 +130,8 @@ func (s *SpanService) checkAndPropose() {
 	if err == nil && lastSpan != nil {
 		nextSpanMsg, err := s.fetchNextSpanDetails(lastSpan.ID+1, lastSpan.EndBlock+1)
 
+		fmt.Println("nextSpanMsg", nextSpanMsg.SelectedProducers)
```
mankenavenkatesh: pls remove this

bridge/pier/syncer.go
```diff
@@ -538,13 +538,16 @@ func (syncer *Syncer) processStateSyncedEvent(eventName string, abiObject *abi.A
...
+			"borChainId", helper.GetConfig().BorChainID,
 		)
 
+		// create clerk event record
 		msg := clerkTypes.NewMsgEventRecord(
 			hmTypes.BytesToHeimdallAddress(helper.GetAddress()),
 			hmTypes.BytesToHeimdallHash(vLog.TxHash.Bytes()),
 			uint64(vLog.Index),
 			event.Id.Uint64(),
+			helper.GetConfig().BorChainID,
```
mankenavenkatesh: FYI - should be added to new bridge implementation as well.

There are few changes to bridge in this PR.

gov/types/codec.go
jdkanani: Added.

bridge/pier/span.go
jdkanani: Removed.

bridge/pier/syncer.go
jdkanani: Yeah. Agree.

cmd/heimdalld/main.go
```diff
@@ -265,7 +265,8 @@ func getGenesisAccount(address []byte) authTypes.GenesisAccount {
 	acc := authTypes.NewBaseAccountWithAddress(hmTypes.BytesToHeimdallAddress(address))
 	genesisBalance, _ := big.NewInt(0).SetString("1000000000000000000000", 10)
 	acc.SetCoins(hmTypes.Coins{hmTypes.Coin{Denom: authTypes.FeeToken, Amount: hmTypes.NewIntFromBigInt(genesisBalance)}})
-	return authTypes.BaseToGenesisAccount(acc)
+	result, _ := authTypes.NewGenesisAccountI(&acc)
```
vaibhavchellani: handle error

gov/client/utils/utils.go
```diff
@@ -0,0 +1,52 @@
...
+//NormalizeProposalType - normalize user specified proposal type
+func NormalizeProposalType(proposalType string) string {
+	switch proposalType {
+	// case "Text", "text":
+	// 	return types.ProposalTypeText
+
+	// case "SoftwareUpgrade", "software_upgrade":
+	// 	return types.ProposalTypeSoftwareUpgrade
+
+	default:
```
vaibhavchellani: Remove comments

gov/endblocker.go
```diff
@@ -0,0 +1,101 @@
...
+		return false
+	})
+
+	// fetch active proposals whose voting periods have ended (are passed the block time)
+	keeper.IterateActiveProposalsQueue(ctx, ctx.BlockHeader().Time, func(proposal types.Proposal) bool {
+		var tagValue, logMsg string
+
+		passes, _, tallyResults := tally(ctx, keeper, proposal)
+
+		// if burnDeposits {
```
vaibhavchellani: Remove?

jdkanani: Done.

gov/client/utils/utils.go
jdkanani: done.

PR #242, opened 2020-03-11
heimdall monitoring

ramakrishna183:
Enabling Prometheus to monitor Heimdall

### DIFF
--- a/monitoring/heimdall_monitoring.sh
+++ b/monitoring/heimdall_monitoring.sh
@@ -14,7 +14,7 @@ json_path=$(pwd)
 echo $json_path
 
 ## Enabling Prometheus for Heimdall
-sudo sed -i 's/prometheus = false/prometheus = true/' /home/ubuntu/.heimdalld/config/config.toml
+sudo sed -i 's/prometheus = false/prometheus = true/' $HOME/.heimdalld/config/config.toml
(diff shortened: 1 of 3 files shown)


monitoring/heimdall_monitoring.sh
```diff
@@ -0,0 +1,96 @@
...
+#            3: Create Grafana Dashboard for Heimdall.		            #
+#									    #
+#############################################################################
+
+## path of Grafana datasorce and dashboard json's
+json_path=$(pwd)
+echo $json_path
+
+## Enabling Prometheus for Heimdall
+sudo sed -i 's/prometheus = false/prometheus = true/' /home/ubuntu/.heimdalld/config/config.toml
```
jdkanani: Can we use the home directory variable `$HOME` or something instead of using `home/ubuntu`?

ramakrishna183: @jdkanani ya we can use **$HOME** or **~/** so, what do you say ?

jdkanani: @ramakrishna183 Sounds good.

vaibhavchellani:
Does this need to be in the Heimdall repo? Also please send PR's to develop first :)

PR #243, opened 2020-03-11
Use tx confirmation time instead of block confirmations...

jdkanani:
... and Heimdall block time instead of the current time

### DIFF
--- a/bridge/pier/syncer.go
+++ b/bridge/pier/syncer.go
@@ -2,6 +2,7 @@ package pier
 
 import (
 	"bytes"
+	"container/list"
 	"context"
 	"encoding/hex"
 	"math/big"
@@ -70,6 +71,9 @@ type Syncer struct {
 
 	// http client to subscribe to
 	httpClient *httpClient.HTTP
+
+	// queue
+	headerQueue *list.List
 }
 
 // NewSyncer returns new service object for syncing events
@@ -104,6 +108,8 @@ func NewSyncer(cdc *codec.Codec, queueConnector *QueueConnector, httpClient *htt
 
 		abis:          abis,
 		HeaderChannel: make(chan *types.Header),
+
+		headerQueue: list.New(),
 	}
 
 	syncer.BaseService = *common.NewBaseService(logger, ChainSyncer, syncer)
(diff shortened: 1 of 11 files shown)


bridge/pier/syncer.go
```diff
@@ -212,17 +224,41 @@ func (syncer *Syncer) startSubscription(ctx context.Context, subscription ethere
...
-	latestNumber := newHeader.Number
+	// adding into queue
+	syncer.headerQueue.PushBack(&LightHeader{
+		Number: newHeader.Number,
+		Time:   newHeader.Time,
+	})
+
+	// current time
+	currentTime := uint64(time.Now().UTC().Unix())
+	confirmationTime := uint64(helper.GetConfig().TxConfirmationTime.Seconds())
```
atvanguard: Can probably keep it in Syncer struct during `NewSyncer` to avoid extra function calls every time.

bridge/pier/syncer.go
```diff
@@ -212,17 +224,41 @@ func (syncer *Syncer) startSubscription(ctx context.Context, subscription ethere
...
+			start = h.Number
+		}
+		end = h.Number
 
-	// confirmation
-	confirmationBlocks := big.NewInt(0).SetUint64(helper.GetConfig().ConfirmationBlocks)
-	confirmationBlocks = confirmationBlocks.Add(confirmationBlocks, big.NewInt(1))
-	if latestNumber.Uint64() > confirmationBlocks.Uint64() {
-		latestNumber = latestNumber.Sub(latestNumber, confirmationBlocks)
+		syncer.headerQueue.Remove(e) // Dequeue
```
atvanguard: Might be a better idea to Dequeue only when we are done processing these batch of blocks.
Say an error occurred during next execution steps, the block won't be re-tried, because it was already dequeued.

bridge/pier/syncer.go
```diff
@@ -234,22 +270,22 @@ func (syncer *Syncer) processHeader(newHeader *types.Header) {
...
-	toBlock := latestNumber
+	toBlock := end
 
 	// debug log
-	syncer.Logger.Debug("Processing header", "fromBlock", fromBlock, "toBlock", toBlock)
+	syncer.Logger.Info("Processing header", "fromBlock", fromBlock, "toBlock", toBlock)
 
 	// set diff
-	if toBlock.Uint64() < fromBlock.Uint64() {
+	if toBlock.Cmp(fromBlock) == -1 {
```
atvanguard: This should never happen. If it does something is wrong in the logic above.

helper/call.go
```diff
@@ -300,7 +301,7 @@ func (c *ContractCaller) IsTxConfirmed(tx common.Hash) bool {
 }
 
 // GetConfirmedTxReceipt returns confirmed tx receipt
-func (c *ContractCaller) GetConfirmedTxReceipt(tx common.Hash) (*ethTypes.Receipt, error) {
+func (c *ContractCaller) GetConfirmedTxReceipt(currentTime time.Time, tx common.Hash) (*ethTypes.Receipt, error) {
 	// get main tx receipt
 	receipt, err := c.GetMainTxReceipt(tx)
 	if err != nil {
```
atvanguard: Should print the error here. It's being ignored in the caller function as well.

bridge/pier/syncer.go
jdkanani: Ah yeah. Agree.

bridge/pier/syncer.go
jdkanani: In new bridge design will have retry mechanism. So, for now, I will keep it for cs-2004.

bridge/pier/syncer.go
jdkanani: Right. Will remove.

PR #269, opened 2020-03-25
Adds --auto-configure for checkpoint

### LINKED ISSUE #266: Add CLI command to send next checkpoint
## Summary

Add CLI command in the `checkpoint` module to send checkpoint without passing external root hash, start/end (optional), account hash.

Reviewers: @jdkanani @mankenavenkatesh 
____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

vaibhavchellani:
Fixes #266 

- Adds a new flag --auto-configure for checkpoint module: This creates the next checkpoint based on the current state of the node and broadcasts it. 

Example command:
```
$ heimdallcli tx checkpoint send-checkpoint --auto-configure=true --chain-id=heimdall-ODAMAV
```

Unfortunately, I had to pass staking keeper to the querier for this one.

### DIFF
--- a/checkpoint/client/cli/tx.go
+++ b/checkpoint/client/cli/tx.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strconv"
 
@@ -47,6 +48,19 @@ func SendCheckpointTx(cdc *codec.Codec) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cliCtx := context.NewCLIContext().WithCodec(cdc)
 			if viper.GetBool(FlagAutoConfigure) {
+				isProposerBytes, _, err := cliCtx.Query(fmt.Sprintf("custom/%s/%s", types.QuerierRoute, types.QueryProposer))
+				if err != nil {
+					return err
+				}
+
+				var isProposer bool
+				if err := json.Unmarshal(isProposerBytes, &isProposer); err != nil {
+					return err
+				}
+				if !isProposer {
+					return errors.New("You are not proposer for this checkpoint, please wait for your turn")
+				}
+
 				// fetch msg checkpoint
 				result, _, err := cliCtx.Query(fmt.Sprintf("custom/%s/%s", types.QuerierRoute, types.QueryNextCheckpoint))
 				if err != nil {
--- a/checkpoint/client/cli/tx.go
+++ b/checkpoint/client/cli/tx.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"encoding/json"
 	"fmt"
 	"strconv"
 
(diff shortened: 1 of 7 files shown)


checkpoint/querier.go
```diff
@@ -113,3 +117,36 @@ func handleQueryCheckpointList(ctx sdk.Context, req abci.RequestQuery, keeper Ke
...
+	rootHash, err := types.GetHeaders(start, end)
+	if err != nil {
+		return nil, sdk.ErrInternal(sdk.AppendMsgToErr(fmt.Sprintf("could not fetch headers for start:%v end:%v error:%v", start, end, err), err.Error()))
+	}
+	accs := sk.GetAllDividendAccounts(ctx)
+	accRootHash, err := types.GetAccountRootHash(accs)
+	if err != nil {
+		return nil, sdk.ErrInternal(sdk.AppendMsgToErr(fmt.Sprintf("could not get generate account root hash. Error:%v", err), err.Error()))
+	}
+	checkpointMsg := types.NewMsgCheckpointBlock(proposer.Signer, start, start+helper.GetConfig().AvgCheckpointLength, hmTypes.BytesToHeimdallHash(rootHash), hmTypes.BytesToHeimdallHash(accRootHash))
```
jdkanani: The proposer should be local address instead of actual proposer otherwise it will at https://github.com/maticnetwork/heimdall/blob/develop/checkpoint/types/msg.go#L55-L57

Or our plan is to allow only the current local address as proposer?

vaibhavchellani: Cool, so if I am not the proposer I will revert for now.

vaibhavchellani:
> Did you test locally?

Yep

jdkanani:
@vaibhavchellani Can you add proposer from local address?

vaibhavchellani:
For this PR I am going to add a return with an error "You are not proposer" if a non proposer sends a checkpoint, because otherwise I would have to add an `from` address in `msg.Checkpoint` which would increase the data being submitted on-chain.

checkpoint/client/cli/tx.go
```diff
@@ -45,6 +46,22 @@ func SendCheckpointTx(cdc *codec.Codec) *cobra.Command {
...
+				}
+
+				// unmarsall the checkpoint msg
+				var newCheckpointMsg types.MsgCheckpoint
+				if err := json.Unmarshal(result, &newCheckpointMsg); err != nil {
+					return err
+				}
+
+				// broadcast this checkpoint
+				return helper.BroadcastMsgsWithCLI(cliCtx, []sdk.Msg{newCheckpointMsg})
```
jdkanani: You can change the `msg.Proposer` just before the broadcast to change the signer.

vaibhavchellani: Non viable: Proposer gets checked in the handler.

vaibhavchellani:
Fixed issue by reverting incase the sender is not proposer. 

Please review and merge @jdkanani @mankenavenkatesh

checkpoint/client/cli/tx.go
```diff
@@ -45,6 +47,35 @@ func SendCheckpointTx(cdc *codec.Codec) *cobra.Command {
 		Short: "send checkpoint to tendermint and ethereum chain ",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cliCtx := context.NewCLIContext().WithCodec(cdc)
+			if viper.GetBool(FlagAutoConfigure) {
+				isProposerBytes, _, err := cliCtx.Query(fmt.Sprintf("custom/%s/%s", types.QuerierRoute, types.QueryProposer))
```
jdkanani: It would be great if we can use `staking/current-proposer` instead of new proposer querier.

vaibhavchellani: I knew there is a querier already, didn't want to pass that here. Will create new obj now from staking keeper.

vaibhavchellani: Yeah that would be the best, missed that.

PR #274, opened 2020-04-01
Signer pubkey 

mankenavenkatesh:


bridge/setu/listener/rootchain.go
```diff
@@ -197,8 +211,12 @@ func (rl *RootChainListener) queryAndBroadcastEvents(fromBlock *big.Int, toBlock
 					if err := helper.UnpackLog(rl.stakingInfoAbi, event, selectedEvent.Name, &vLog); err != nil {
 						rl.Logger.Error("Error while parsing event", "name", selectedEvent.Name, "error", err)
 					}
-					if bytes.Compare(event.NewSigner.Bytes(), helper.GetAddress()) == 0 {
+					if bytes.Compare(event.NewSignerPubkey, helper.GetPubKey().Bytes()) == 0 {
```
jdkanani, line 214: Event's public-key won't have `0x04` prefix while `helper.GetPubkey()` has a prefix. So it won't match directly. Also added some comments to add `signer` in the event: https://github.com/maticnetwork/contracts/pull/216#pullrequestreview-385289830

You might have to change ABI for the same.

mankenavenkatesh: oh yeah....Great catch. thank you.

bridge/setu/util/common.go
```diff
@@ -315,3 +315,13 @@ func GetAccount(cliCtx cliContext.CLIContext) (account authTypes.Account, err er
...
 	return
 }
+
+// appendPrefix - returns publickey in uncompressed format
+func AppendPrefix(signerPubKey []byte) []byte {
+	// append prefix - "0x04" as heimdall uses publickey in uncompressed format. Refer below link
+	// https://superuser.com/questions/1465455/what-is-the-size-of-public-key-for-ecdsa-spec256r1
+	prefix := make([]byte, 1)
+	prefix[0] = byte(0x04)
+	signerPubKey = append(prefix[:], signerPubKey[:]...)
```
jdkanani, line 325: `[]byte{0x40}`

PR #275, opened 2020-04-02
prometheus docker for heimdall monitoring 

ramakrishna183:


monitoring/docker/provisioning/dashboards/dashboards.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: 1
+
+providers:
+- name: 'a unique provider name'
```
mankenavenkatesh, line 4: can we add Provider name?

mankenavenkatesh:
Please add README providing details on how to start 
1. how to start both the docker containers.
2. URL to access dashboard.
3. Any configurations to be done on dashboard 
4. What are different metrics available. Share any links if available.

monitoring/docker/provisioning/dashboards/dashboards.yaml
jdkanani: @mankenavenkatesh Please suggest a name for the same.

PR #285, opened 2020-04-09
fix: use make install while creating linux package

jdkanani:


.github/workflows/linuxpackage.yml
```diff
@@ -31,7 +31,10 @@ jobs:
           gem install --no-document fpm
           fpm --version
 
-          make build
+          make install
+          mkdir -p build
+          export GOBIN=$(go env GOPATH)/bin
```
mankenavenkatesh, line 36: why is this required?

PR #294, opened 2020-04-12
Auth module testcases

jdkanani:


app/app.go
```diff
@@ -458,6 +459,15 @@ func (app *HeimdallApp) InitChainer(ctx sdk.Context, req abci.RequestInitChain)
 	if err := ModuleBasics.ValidateGenesis(genesisState); err != nil {
 		panic(err)
 	}
+
+	// check fee collector module account
+	if moduleAcc := app.SupplyKeeper.GetModuleAccount(ctx, authTypes.FeeCollectorName); moduleAcc == nil {
+		if moduleAcc == nil {
```
temaniarpit27, line 465: same as prev condition

auth/ante_test.go
```diff
@@ -0,0 +1,367 @@
...
+	require.True(sdk.IntEq(t, happ.SupplyKeeper.GetModuleAccount(ctx, types.FeeCollectorName).GetCoins().AmountOf(authTypes.FeeToken), amt))
+	require.True(sdk.IntEq(t, happ.AccountKeeper.GetAccount(ctx, hmTypes.AccAddressToHeimdallAddress(addr1)).GetCoins().AmountOf(authTypes.FeeToken), sdk.NewInt(0)))
+}
+
+//
+// utils
+//
+
+// run the tx through the anteHandler and ensure its valid
+func checkValidTx(t *testing.T, anteHandler sdk.AnteHandler, ctx sdk.Context, tx sdk.Tx, simulate bool) (sdk.Context, sdk.Result, bool) {
```
tokamak-git, line 340: too many inputs to func, should we use a struct instead? or make this a method on an existing struct?

auth/ante_test.go
```diff
@@ -0,0 +1,367 @@
...
+	var tx sdk.Tx
+	msg1 := sdkAuth.NewTestMsg(addr1)
+	msg2 := sdkAuth.NewTestMsg(addr2)
+
+	acc1 = happ.AccountKeeper.GetAccount(ctx, hmTypes.AccAddressToHeimdallAddress(addr1))
+	acc2 = happ.AccountKeeper.GetAccount(ctx, hmTypes.AccAddressToHeimdallAddress(addr2))
+	// accNumber1 := acc1.GetAccountNumber()
+	// accNumber2 := acc2.GetAccountNumber()
+
+	// test good tx from one signer
```
tokamak-git, line 231: Should we use a loop for these tests?
```
create slice of struct of test cases
run them via loop
pass them to required functions
```

auth/ante_test.go
```diff
@@ -0,0 +1,367 @@
...
+	newCtx, result, abort := anteHandler(ctx, tx, simulate)
+	require.Equal(t, "", result.Log)
+	require.False(t, abort)
+	require.Equal(t, sdk.CodeOK, result.Code)
+	require.True(t, result.IsOK())
+	return newCtx, result, abort
+}
+
+// run the tx through the anteHandler and ensure it fails with the given code
+func checkInvalidTx(t *testing.T, anteHandler sdk.AnteHandler, ctx sdk.Context, tx sdk.Tx, simulate bool, code sdk.CodeType) (sdk.Context, sdk.Result, bool) {
```
tokamak-git, line 350: same as above, a transaction struct can have `valid` and `invalid` trx methods. what do you think?

auth/ante_test.go
```diff
@@ -0,0 +1,367 @@
...
+	tx = types.NewTestTx(ctx, msg1, priv1, accNumber1, uint64(0))
+	checkValidTx(t, anteHandler, ctx, tx, false)
+
+	// test good tx from one signer
+	tx = types.NewTestTx(ctx, msg1, priv1, accNumber1, uint64(1))
+	checkValidTx(t, anteHandler, ctx, tx, false)
+
+	// // new tx from wrong account number
+	tx = types.NewTestTx(ctx, msg2, priv2, accNumber2, uint64(1))
+	checkInvalidTx(t, anteHandler, ctx, tx, false, sdk.CodeUnauthorized)
```
tokamak-git, line 289: Dont we want to check for a specific error?

auth/ante_test.go
```diff
@@ -0,0 +1,367 @@
...
+	priv1, _, addr1 := sdkAuth.KeyTestPubAddr()
+	msg1 := sdkAuth.NewTestMsg(addr1)
+	tx1 := types.NewTestTx(ctx, msg1, priv1, uint64(0), uint64(0)) // use sdk's auth module for msg
+
+	_, result1, _ := checkInvalidTx(t, anteHandler, ctx, tx1, false, sdk.CodeInternal)
+	require.Contains(t, result1.Log, "fee_collector module account has not been set")
+}
+
+func (suite *AnteTestSuite) TestGasLimit() {
+	t, happ, ctx, anteHandler := suite.T(), suite.app, suite.ctx, suite.anteHandler
```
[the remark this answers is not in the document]
jdkanani: Basically, we are testing the gas limit at height 1.

auth/ante_test.go
```diff
@@ -0,0 +1,367 @@
...
+// Tests
+//
+
+func (suite *AnteTestSuite) TestAnteValidation() {
+	t, happ, ctx, anteHandler := suite.T(), suite.app, suite.ctx, suite.anteHandler
+
+	// module account validation
+	// removing module account
+	acc := happ.SupplyKeeper.GetModuleAccount(ctx, authTypes.FeeCollectorName)
+	happ.AccountKeeper.RemoveAccount(ctx, acc)
```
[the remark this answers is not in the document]
jdkanani: Yeah, we can. But easier to read I guess.

auth/ante_test.go
```diff
@@ -0,0 +1,367 @@
...
+func TestAnteTestSuite(t *testing.T) {
+	suite.Run(t, new(AnteTestSuite))
+}
+
+//
+// Tests
+//
+
+func (suite *AnteTestSuite) TestAnteValidation() {
+	t, happ, ctx, anteHandler := suite.T(), suite.app, suite.ctx, suite.anteHandler
```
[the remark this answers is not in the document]
jdkanani: One use directly `suite.T()`, `suite.app` directly -- just doing it for clean code.

auth/ante_test.go
jdkanani: No need to. This is just an internal function. But yeah we can create a struct.

auth/ante_test.go
jdkanani: Yeah, better suggestion. We should do that. We can use `t.Run` as well.

auth/ante_test.go
jdkanani: `checkInvalidTx` is doing the same. It will validate `sdk. CodeUnauthorized` error.

auth/ante_test.go
tokamak-git: Was hoping there is an existing transaction struct with the same functionality that could have been used. So sure makes sense to keep it the way it is.

auth/ante_test.go
tokamak-git: resolved same as query on L340.

PR #296, opened 2020-04-13
mat-1039 Check if signer are available

ptsayli:


auth/ante.go
```diff
@@ -145,6 +145,11 @@ func NewAnteHandler(
 		// stdSigs contains the sequence number, account number, and signatures.
 		// When simulating, this would just be a 0-length slice.
 		signerAddrs := stdTx.GetSigners()
+
+		if len(signerAddrs) <= 0 {
```
vaibhavchellani, line 149: Will it be `less than` ever?

jdkanani: @ptsayli

PR #297, opened 2020-04-13
mat-1037-check txbytes length while pulp decoding

ptsayli:


auth/types/pulp.go
```diff
@@ -73,6 +74,11 @@ func (p *Pulp) EncodeToBytes(tx StdTx) ([]byte, error) {
 // DecodeBytes decodes bytes to msg
 func (p *Pulp) DecodeBytes(data []byte) (interface{}, error) {
 	var txRaw StdTxRaw
+
+	if len(data) <= PulpHashLength {
+		return nil, errors.New("Invalid data size")
```
vaibhavchellani, line 79:
```suggestion
		return nil, errors.New("Invalid data length, should be greater than PulpPrefix")
```

ptsayli:
https://github.com/maticnetwork/heimdall/issues/289

PR #300, opened 2020-04-15
MAT-1088 only allow one signer in tx and remove multi-send

jdkanani:


auth/ante.go
```diff
@@ -150,6 +150,10 @@ func NewAnteHandler(
 			return newCtx, sdk.ErrUnauthorized("no signer exists").Result(), true
 		}
 
+		if len(signerAddrs) > 1 {
+			return newCtx, sdk.ErrUnauthorized("wrong number of signers").Result(), true
+		}
+
```
mankenavenkatesh, line 156: At line no 190, we don't need a for loop i think because of this change. 
		
for i := 0; i < len(stdSigs); i++ {
			// skip the fee payer, account is cached and fees were deducted already
			if i != 0 {
				signerAccs[i], res = GetSignerAcc(newCtx, ak, types.AccAddressToHeimdallAddress(signerAddrs[i]))
				if !res.IsOK() {
					return newCtx, res, true
				}
			}

jdkanani: Fixed.

PR #302, opened 2020-04-17
mat-1052- Topup module testcases

ptsayli:


topup/handler_test.go
```diff
@@ -0,0 +1,145 @@
...
+	querier        sdk.Querier
+	handler        sdk.Handler
+	contractCaller mocks.IContractCaller
+}
+
+// SetupTest setup all necessary things for querier tesing
+func (suite *HandlerTestSuite) SetupTest() {
+	suite.app, suite.ctx, suite.cliCtx = createTestApp(false)
+	suite.contractCaller = mocks.IContractCaller{}
+	suite.querier = topup.NewQuerier(suite.app.TopupKeeper, &suite.contractCaller)
```
jdkanani, line 42: No need to create querier to test handler.

ptsayli: my bad 👍

PR #307, opened 2020-04-22
MAT-1135: adding checks for secret files

tokamak-git:
adding check for secret files. it shall log an error if permissions are not set as expected.
added a file permission checker func. it is called in the bridge start.


tokamak-git:
addressed

file/permCheck.go
```diff
@@ -0,0 +1,31 @@
+package file
```
jdkanani, line 1: On file names, shouldn't we use `_` or `-` separation for file names?

tokamak-git: using snake case for file names

types/error/errInvalidPermissions.go
```diff
@@ -0,0 +1,13 @@
+package types
```
jdkanani, line 1: Please rename filename to `invalid_permissions.go`, just to follow filename conventions.

tokamak-git:
@mankenavenkatesh hey please let me know if I can go ahead with merging this. have implemented the changes you requested.

PR #309, opened 2020-04-22
MAT-1164: audit: improvement: modified getPulpHash to accept sdk.Msg

tokamak-git:
modified getPulpHash to accept sdk.Msg
ref: `https://linear.app/issue/MAT-1164/pulp-getpulphash-should-take-sdkmsg-instead-of-string`


auth/types/pulp_test.go
```diff
@@ -0,0 +1,20 @@
+package types
+
+import (
+	"testing"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"gotest.tools/assert"
```
jdkanani, line 7: Let's use `testify` since we are using it for other test cases as well.

tokamak-git: updated

tokamak-git: @jdkanani made changes as requested.

PR #313, opened 2020-04-23
mat-1161-stdxvalidatebasic should validate Tx

ptsayli:


jdkanani:
@ptsayli Please add test cases for the same.

PR #318, opened 2020-04-29
Clerk Package Test Cases (mat-1224)

temaniarpit27:


clerk/handler_test.go
```diff
@@ -0,0 +1,171 @@
...
+	// 	txHash,
+	// 	logIndex,
+	// 	id,
+	// 	chainId,
+	// )
+	// eventLog = &statesender.StatesenderStateSynced{
+	// 	Id:              big.NewInt(5),
+	// 	ContractAddress: address,
+	// }
+	// suite.contractCaller.On("GetConfirmedTxReceipt", mock.Anything, txHash.EthHash(), chainParams.TxConfirmationTime).Return(nil, errors.New("txn receipt error"))
```
temaniarpit27, line 166: @mankenavenkatesh : not sure why but this test case is passing whereas it should fail i think
because i am returning error from the contractCaller

clerk/handler_test.go
```diff
@@ -0,0 +1,87 @@
...
+	suite.app, suite.ctx = createTestApp(false)
+	suite.contractCaller = mocks.IContractCaller{}
+	suite.handler = clerk.NewHandler(suite.app.ClerkKeeper, &suite.contractCaller)
+}
+
+func TestHandlerTestSuite(t *testing.T) {
+	suite.Run(t, new(HandlerTestSuite))
+}
+
+func (suite *HandlerTestSuite) TestHandleMsgEventRecord() {
```
tokamak-git, line 45: can this be written as a set of cases? it would reduce duplication of code.

clerk/keeper_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+func TestKeeperTestSuite(t *testing.T) {
+	suite.Run(t, new(KeeperTestSuite))
+}
+
+//
+// Tests
+//
+
+func (suite *KeeperTestSuite) TestHasGetSetEventRecord() {
```
tokamak-git, line 40: should these be broken into separate tests?

clerk/keeper_test.go
```diff
@@ -0,0 +1,124 @@
...
+func (suite *KeeperTestSuite) TestSetHasGetRecordSequence() {
+	t, app, ctx := suite.T(), suite.app, suite.ctx
+
+	testSeq := "testseq"
+	ck := app.ClerkKeeper
+	ck.SetRecordSequence(ctx, testSeq)
+	found := ck.HasRecordSequence(ctx, testSeq)
+	require.True(t, found)
+
+	found = ck.HasRecordSequence(ctx, "testSeq")
```
tokamak-git, line 119: is this not a repetition of the test? the `testSeq` var contains a string rite? not clear as to what is being tested...

app/test_helpers.go
```diff
@@ -349,6 +350,34 @@ func SetupTopupGenesis() *HeimdallApp {
 	return app
 }
 
+// SetupClerkGenesis initializes a new Heimdall with the default genesis data.
+func SetupClerkGenesis() *HeimdallApp {
```
ptsayli, line 354: we can move/merge this to `clerk/integration_test.go/createTestApp`

clerk/querier.go
```diff
@@ -15,22 +15,22 @@ import (
...
-			return handleQueryRecordSequence(ctx, req, keeper)
+			return handleQueryRecordSequence(ctx, req, keeper, contractCaller)
 		default:
 			return nil, sdk.ErrUnknownRequest("unknown auth query endpoint")
 		}
 	}
 }
 
-func handleQueryRecord(ctx sdk.Context, req abci.RequestQuery, keeper Keeper) ([]byte, sdk.Error) {
+func handleQueryRecord(ctx sdk.Context, req abci.RequestQuery, keeper Keeper, contractCallerObj helper.IContractCaller) ([]byte, sdk.Error) {
```
ptsayli, line 33: `contractCallerObj` is not used, can remove

clerk/querier.go
```diff
@@ -55,7 +50,7 @@ func handleQueryRecord(ctx sdk.Context, req abci.RequestQuery, keeper Keeper) ([
 	return bz, nil
 }
 
-func handleQueryRecordList(ctx sdk.Context, req abci.RequestQuery, keeper Keeper) ([]byte, sdk.Error) {
+func handleQueryRecordList(ctx sdk.Context, req abci.RequestQuery, keeper Keeper, contractCallerObj helper.IContractCaller) ([]byte, sdk.Error) {
```
ptsayli, line 53: please remove `contractCallerObj`

clerk/keeper_test.go
temaniarpit27: the values are different
testseq
testSeq

clerk/keeper_test.go
temaniarpit27: i think we should continue with this
getting an event record will involve positive and negative test case - finding the record and not finding
since we have already set once, we can use that to test other cases

clerk/handler_test.go
temaniarpit27: same as below

temaniarpit27:
@ptsayli @tokamak-git : made changes and left comments

PR #321, opened 2020-05-01
mat-1121 add simulation for staking module

ptsayli:
Package: app
coverage: 49.1% of statements

### DIFF
--- a/app/sim_test.go
+++ b/app/sim_test.go
@@ -202,20 +202,19 @@ func TestAppSimulationAfterImport(t *testing.T) {
 		require.NoError(t, os.RemoveAll(newDir))
 	}()
 
-	newApp := NewHeimdallApp(logger, db)
+	newApp := NewHeimdallApp(logger, newDB)
 	require.Equal(t, AppName, app.Name())
 
 	newApp.InitChain(abci.RequestInitChain{
 		AppStateBytes: appState,
 	})
-
-	require.Panics(t, func() {
-		simulation.SimulateFromSeed(
-			t, os.Stdout, newApp.BaseApp, AppStateFn(app.Codec(), app.SimulationManager()),
-			SimulationOperations(newApp, newApp.Codec(), config),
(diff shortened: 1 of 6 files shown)


app/sim_test.go
```diff
@@ -204,22 +202,20 @@ func TestAppSimulationAfterImport(t *testing.T) {
...
-		newApp.ModuleAccountAddrs(), config,
-	)
-	require.NoError(t, err)
+	require.Panics(t, func() {
+		simulation.SimulateFromSeed(
+			t, os.Stdout, newApp.BaseApp, AppStateFn(app.Codec(), app.SimulationManager()),
+			SimulationOperations(newApp, newApp.Codec(), config),
+			newApp.ModuleAccountAddrs(), config,
+		)
+	})
```
jdkanani, line 218: This shouldn't panic!

ptsayli: We have set block limit` to 500, and that is already imported from `previous State`
so it panic here
```
prevHeight := app.LastBlockHeight()
if req.Header.Height != prevHeight+1 {
	return fmt.Errorf("invalid height: %d; expected: %d", req.Header.Height, prevHeight+1)
}
```

jdkanani: Maybe we are using old DB. Block should start from 0 after import, but the init state should be old.

app/sim_test.go
```diff
@@ -204,22 +202,20 @@ func TestAppSimulationAfterImport(t *testing.T) {
 		require.NoError(t, os.RemoveAll(newDir))
 	}()
 
-	newApp := NewHeimdallApp(logger, db, nil)
-	if err := newApp.LoadLatestVersion(newApp.keys[bam.MainStoreKey]); err != nil {
-		require.NoError(t, err)
-	}
+	newApp := NewHeimdallApp(logger, db)
```
jdkanani, line 205:
```suggestion
	newApp := NewHeimdallApp(logger, newDB)
```

PR #322, opened 2020-05-03
Sidechannel span

mankenavenkatesh:


bor/handler.go
```diff
@@ -44,43 +47,29 @@ func HandleMsgProposeSpan(ctx sdk.Context, msg types.MsgProposeSpan, k Keeper) s
...
-	}
-
-	// get last span
-	_, err = k.GetLastSpan(ctx)
-	if err != nil {
-		k.Logger(ctx).Error("Unable to fetch last span", "Error", err)
-		return common.ErrSpanNotFound(k.Codespace()).Result()
+	// Validate Span duration
+	spanDuration := k.GetParams(ctx).SpanDuration
+	if spanDuration != (msg.EndBlock - msg.StartBlock) {
```
jdkanani, line 65: Shouldn't it be as followed:

```suggestion
	if spanDuration != (msg.EndBlock - msg.StartBlock + 1) {
```

bor/client/rest/query.go
```diff
@@ -245,7 +287,7 @@ func prepareNextSpanHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
 		msg := hmTypes.NewSpan(
 			spanID,
 			startBlock,
-			startBlock+spanDuration-1,
+			startBlock+spanDuration,
```
jdkanani, line 290:
```suggestion
			startBlock + spanDuration - 1,
```

Example: Let's span duration is `6400` and `startBlock` is `0`. Then `endBlock` should be:

```
endBlock := startBlock + spanDuration - 1 = 0 + 6400 - 1 = 6399
```

So, now span will start at block `0` and end at `6399`, total of `6400`

bor/client/rest/query.go
```diff
@@ -22,9 +22,51 @@ func registerQueryRoutes(cliCtx context.CLIContext, r *mux.Router) {
...
+			var nextSpanSeed = common.BytesToHash(res)
+			RestLogger.Debug("Fetched next span seed ", "NextSpanSeed", nextSpanSeed)
+
+			result, err := json.Marshal(&nextSpanSeed)
+			if err != nil {
+				RestLogger.Error("Error while marshalling response to Json", "error", err)
+				rest.WriteErrorResponse(w, http.StatusBadRequest, err.Error())
+				return
+			} */
+
```
jdkanani, line 63: Remove unwanted comments.

PR #324, opened 2020-05-04
MAT-1253: testing: added base app constructor for bor module keeper t…

tokamak-git:
…esting


tokamak-git:
going ahead with #332 instead

PR #325, opened 2020-05-04
Fix Bridge Logger Issue (mat-1316)

temaniarpit27:


bridge/setu/listener/service.go
```diff
@@ -16,7 +16,7 @@ const (
 	MaticChainListenerStr = "maticchain"
 )
 
-var logger = util.Logger().With("service", ListenerServiceStr)
+// var logger = util.Logger().With("service", ListenerServiceStr)
```
mankenavenkatesh, line 19: remove this

PR #326, opened 2020-05-05
Fix bridge logger

mankenavenkatesh:


bridge/setu/listener/service.go
```diff
@@ -17,7 +17,7 @@ const (
 	MaticChainListenerStr = "maticchain"
 )
 
-var logger = util.Logger().With("service", ListenerServiceStr)
+// var logger = util.Logger().With("service", ListenerServiceStr)
```
jdkanani, line 20: Remove unwanted comments.

PR #332, opened 2020-05-07
MAT-1253: testing: bor module keeper testing testing freeze function

tokamak-git:


merr/validation_err.go
```diff
@@ -0,0 +1,12 @@
+package merr
```
mankenavenkatesh, line 1: Can we add errors to package `common`

bor/keeper_test.go
```diff
@@ -0,0 +1,69 @@
+package bor_test
+
+import (
+	"fmt"
+	"testing"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/magiconair/properties/assert"
```
mankenavenkatesh, line 8: Let's use assert from `stretchr` package which is commonly used by other test cases

go.mod
```diff
@@ -27,6 +27,7 @@ require (
 	github.com/influxdata/influxdb v1.2.3-0.20171009172446-a55dd0f50edd // indirect
 	github.com/jackpal/go-nat-pmp v1.0.2 // indirect
 	github.com/karalabe/usb v0.0.0-20191104083709-911d15fe12a9 // indirect
+	github.com/magiconair/properties v1.8.1
```
mankenavenkatesh, line 30: This is not required.

app/app.go
```diff
@@ -471,6 +471,20 @@ func MakeCodec() *codec.Codec {
...
 
+// MakePulp creates pulp codec and registers custom types for decoder
+func MakePulp() *authTypes.Pulp {
+	pulp := authTypes.GetPulpInstance()
+
+	// register custom type
+	checkpointTypes.RegisterPulp(pulp)
+
+	return pulp
+}
```
jdkanani, line 482: Pulp is removed in sidechannel

merr/validation_err.go
tokamak-git: kept it as an independent package so as to allow for future addition of additional errors. also becomes a logical separation and would be easy for anyone to look at one place for all error related stuff, also so it can be accessed by linters and language servers appropiately.

app/app.go
tokamak-git: removed

go.mod
tokamak-git: removed

bor/keeper_test.go
tokamak-git: removed

types/span.go
```diff
@@ -46,3 +46,11 @@ func SortSpanByID(a []*Span) {
...
 	})
 }
+
+// Valid is invoked to check validity of the span
+func (s *Span) Valid() (valid bool) {
+	if s.ID == 0 {
+		return
+	}
+	return true
+}
```
jdkanani, line 56: Span with `ID=0` is also valid. That's the initial span embedded into genesis file.

tokamak-git: If I send id as zero, it panics. thats coz it fails genesis validation?

app/app.go
```diff
@@ -471,6 +471,10 @@ func MakeCodec() *codec.Codec {
 	return cdc
 }
 
+func (app *HeimdallApp) GetCaller() helper.ContractCaller {
```
ptsayli, line 474: Is this really required,
why can't we directly use `app.caller`?

tokamak-git: the field is private to the struct, thus we cannot access it in the tests as they are running in `bor_test` package

tokamak-git:
merged to develop with #365

PR #339, opened 2020-05-09
MAT-1132 - Sidechannel confirmations

mankenavenkatesh:
Use confirmation blocks instead of confirmation time

This PR has fixes for
- https://linear.app/issue/MAT-1132/use-confirmation-blocks-instead-of-confirmation-time
- https://linear.app/issue/MAT-1305/10-bor-blocks-as-confirmation-for-checkpoint

### DIFF
--- a/bridge/setu/listener/maticchain.go
+++ b/bridge/setu/listener/maticchain.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/RichardKnop/machinery/v1/tasks"
 	"github.com/maticnetwork/bor/core/types"
-	"github.com/maticnetwork/heimdall/bridge/setu/util"
 	"github.com/maticnetwork/heimdall/helper"
 )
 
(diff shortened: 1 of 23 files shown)


bridge/setu/listener/maticchain.go
```diff
@@ -62,14 +61,7 @@ func (ml *MaticChainListener) ProcessHeader(newHeader *types.Header) {
...
 
-	chainmanagerParams, err := util.GetChainmanagerParams(ml.cliCtx)
-	if err != nil {
-		ml.Logger.Error("Error fetching chain manager params", "error", err)
-		return
-	}
-
-	confirmationTime := chainmanagerParams.TxConfirmationTime
-	ml.sendTaskWithDelay("sendCheckpointToHeimdall", headerBytes, confirmationTime)
+	ml.sendTaskWithDelay("sendCheckpointToHeimdall", headerBytes, 0)
```
ptsayli, line 64: Don't you think we should keep delay
otherwise, it will unnecessarily attempt to send checkpoint 
TimeConfirmation + BlockConfirmation check would be better

mankenavenkatesh: We are replacing timeconfirmation with blockConfirmations. 
Added block confirmations check on processor side which process only confirmed blocks.

bridge/setu/listener/rootchain.go
```diff
@@ -95,42 +91,24 @@ func (rl *RootChainListener) Start() error {
...
-	}
-
-	if start == nil {
-		return
+	requiredConfirmations := rootchainContext.ChainmanagerParams.MainchainTxConfirmations
+	latestNumber := newHeader.Number
+
+	// confirmation
+	confirmationBlocks := big.NewInt(0).SetUint64(requiredConfirmations)
+	confirmationBlocks = confirmationBlocks.Add(confirmationBlocks, big.NewInt(1))
```
jdkanani, line 104: Do we need this line?

```suggestion
```

bridge/setu/listener/rootchain.go
```diff
@@ -95,42 +91,24 @@ func (rl *RootChainListener) Start() error {
...
+	if latestNumber.Uint64() > confirmationBlocks.Uint64() {
+		latestNumber = latestNumber.Sub(latestNumber, confirmationBlocks)
```
jdkanani, line 106:
```suggestion
	if latestNumber.Cmp(confirmationBlocks) <= 0 {
		return
	}
	latestNumber = latestNumber.Sub(latestNumber, confirmationBlocks)
```

PR #349, opened 2020-05-13
Add more clerk testcases

jdkanani:


jdkanani:
@mankenavenkatesh Wouldn't set sequence test case covers it?

mankenavenkatesh:
There is no sequence check in `PostHandleMsgEventRecord`. 
It's present only in `handleMsgEventRecord`.

PR #360, opened 2020-05-18
Add message for build status (mat-1448)

temaniarpit27:


Makefile
```diff
@@ -29,6 +29,7 @@ build: clean
 	go build -o build/heimdalld ./cmd/heimdalld
 	go build -o build/heimdallcli ./cmd/heimdallcli
 	go build -o build/bridge bridge/bridge.go
+	echo "====================================================\n==================Build Successful==================\n===================================================="
```
mankenavenkatesh, line 32: we should check `exit status` of `build` command.

PR #361, opened 2020-05-18
MAT-1449 - Remove slash limit

mankenavenkatesh:
Tick should be submitted to heimdall even if slash-limit is not exceeded.

### DIFF
--- a/slashing/handler.go
+++ b/slashing/handler.go
@@ -3,7 +3,6 @@ package slashing
 import (
 	"bytes"
 	"encoding/hex"
-	"fmt"
 	"math/big"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
@@ -51,9 +50,9 @@ func handlerMsgTick(ctx sdk.Context, msg types.MsgTick, k Keeper, contractCaller
 
 	// check if slash limit is exceeded or not
 	totalSlashedAmount := k.GetTotalSlashedAmount(ctx)
-	if totalSlashedAmount <= 0 {
-		k.Logger(ctx).Error("TotalSlashedAmount is less than or equal to zero")
-		return hmCommon.ErrInvalidMsg(k.Codespace(), fmt.Sprintf("TotalSlashedAmount is less than or equal to zero", totalSlashedAmount)).Result()
+	if totalSlashedAmount == 0 {
+		k.Logger(ctx).Error("Slashed amount is zero")
+		return hmCommon.ErrInvalidMsg(k.Codespace(), "Slashed amount is zero", totalSlashedAmount).Result()
 	}
 
 	// check if tick msgs are in continuity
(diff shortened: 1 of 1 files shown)


slashing/handler.go
```diff
@@ -50,9 +50,10 @@ func handlerMsgTick(ctx sdk.Context, msg types.MsgTick, k Keeper, contractCaller
 	)
 
 	// check if slash limit is exceeded or not
-	if !k.IsSlashedLimitExceeded(ctx) {
-		k.Logger(ctx).Error("TotalSlashedAmount is less than SlashLimit")
-		return hmCommon.ErrInvalidMsg(k.Codespace(), fmt.Sprintf("TotalSlashedAmount %v is less than SlashLimit", k.GetTotalSlashedAmount(ctx))).Result()
+	totalSlashedAmount := k.GetTotalSlashedAmount(ctx)
+	if totalSlashedAmount <= 0 {
```
jdkanani, line 54: `GetTotalSlashedAmount` returns uint64`, so it will never be < 0.

```suggestion
	if totalSlashedAmount == 0 {
```

slashing/handler.go
```diff
@@ -50,9 +50,10 @@ func handlerMsgTick(ctx sdk.Context, msg types.MsgTick, k Keeper, contractCaller
...
+		k.Logger(ctx).Error("TotalSlashedAmount is less than or equal to zero")
```
jdkanani, line 55:
```suggestion
		k.Logger(ctx).Error("Slashed amount is zero")
```

slashing/handler.go
```diff
@@ -50,9 +50,10 @@ func handlerMsgTick(ctx sdk.Context, msg types.MsgTick, k Keeper, contractCaller
...
+		return hmCommon.ErrInvalidMsg(k.Codespace(), fmt.Sprintf("TotalSlashedAmount is less than or equal to zero", totalSlashedAmount)).Result()
```
jdkanani, line 56:
```suggestion
		return hmCommon.ErrInvalidMsg(k.Codespace(), "Slashed amount is zero").Result()
```

slashing/handler.go
```diff
@@ -50,9 +49,10 @@ func handlerMsgTick(ctx sdk.Context, msg types.MsgTick, k Keeper, contractCaller
 	)
 
 	// check if slash limit is exceeded or not
-	if !k.IsSlashedLimitExceeded(ctx) {
-		k.Logger(ctx).Error("TotalSlashedAmount is less than SlashLimit")
-		return hmCommon.ErrInvalidMsg(k.Codespace(), fmt.Sprintf("TotalSlashedAmount %v is less than SlashLimit", k.GetTotalSlashedAmount(ctx))).Result()
+	totalSlashedAmount := k.GetTotalSlashedAmount(ctx)
+	if totalSlashedAmount == 0 {
+		k.Logger(ctx).Error("Slashed amount is zero")
+		return hmCommon.ErrInvalidMsg(k.Codespace(), "Slashed amount is zero", totalSlashedAmount).Result()
```
jdkanani, line 55:
```suggestion
		return hmCommon.ErrInvalidMsg(k.Codespace(), "Slashed amount is zero").Result()
```

PR #379, opened 2020-05-25
mat-1478 Checkpoint ack sent by multiple validators

ptsayli:


bridge/setu/processor/checkpoint.go
```diff
@@ -265,9 +265,12 @@ func (cp *CheckpointProcessor) sendCheckpointAckToHeimdall(eventName string, che
 			"txHash", hmTypes.BytesToHeimdallHash(log.TxHash.Bytes()),
 			"logIndex", uint64(log.Index),
 		)
+		latestCheckpoint, err := util.GetlastestCheckpoint(cp.cliCtx)
 
-		// TODO - check if this ack is already processed on heimdall or not.
-		// TODO - check if i am the proposer of this ack or not.
+		if err == nil && latestCheckpoint.StartBlock == event.Start.Uint64() && latestCheckpoint.EndBlock == event.End.Uint64() {
```
mankenavenkatesh, line 270:
```suggestion
		if err == nil && latestCheckpoint  != nil && latestCheckpoint.StartBlock == event.Start.Uint64() && latestCheckpoint.EndBlock == event.End.Uint64() {
```

PR #383, opened 2020-05-30
Bor simulation test

ptsayli:


bor/genesis_test.go
```diff
@@ -0,0 +1,67 @@
...
+func (suite *GenesisTestSuite) TestInitExportGenesis() {
+	t, app, ctx := suite.T(), suite.app, suite.ctx
+	spancount := 5
+	params := types.DefaultParams()
+	spans := []*hmTypes.Span{}
+	chainID := "15001"
+	start := uint64(0)
+	end := uint64(0)
+	chSim.LoadValidatorSet(4, t, app.StakingKeeper, ctx, false, 10)
+	app.BorKeeper.SetParams(ctx, types.DefaultParams())
```
tokamak-git, line 42: `params` holds the same values correct, any reason not to reuse?

bor/genesis_test.go
```diff
@@ -0,0 +1,67 @@
...
+	valSet := app.StakingKeeper.GetValidatorSet(ctx)
```
tokamak-git, line 43: `LoadValidatorSet` above returns the validators set, any reason we cant use that directly?

bor/genesis_test.go
```diff
@@ -0,0 +1,67 @@
...
+		)
+		spans = append(spans, &span)
+	}
+	genesisState := types.NewGenesisState(
+		params,
+		spans,
+	)
+	bor.InitGenesis(ctx, app.BorKeeper, genesisState)
+	actualParams := bor.ExportGenesis(ctx, app.BorKeeper)
+	require.NotNil(t, actualParams)
```
tokamak-git, line 65: dont we want to compare the value being returned?

bor/genesis_test.go
ptsayli: my bad : 🤗

bor/genesis_test.go
ptsayli: can, but imported and exported genesis will not be exact same
so let's keep this as it is

PR #388, opened 2020-06-16
MAT-1477 - Bridge optimizations

mankenavenkatesh:
This PR contains following bridge optimisations
```
MAT-1696 - Multiple span transactions for same spanID

Solution
- Add delay between each proposer.
```

```
MAT-1673 - Retry `tx` from bridge if failed during side-tx-processing

Solution
- After broadcasting transaction from bridge, add back the msg to queue with retry delay.
`return tasks.NewErrRetryTaskLater("some error", 6 * estimatedNextBlockTime)`

- After `6 blocks` produced on heimdall, same msg will be retried by bridge.
- As we validate and check if msg has been already processed or not on heimdall using `sequenceNo`,
    - if already processed, bridge will ignore
    - else, bridge will send the msg again repeating the process.
```

```
MAT-1228 - Add context params to cache to reduce params api call
Solution
- Cache the params and  reload at frequent intervals. (1hr)
```

### DIFF
--- a/bridge/setu/listener/maticchain.go
+++ b/bridge/setu/listener/maticchain.go
@@ -8,11 +8,13 @@ import (
 	"github.com/maticnetwork/bor/core/types"
 	"github.com/maticnetwork/heimdall/bridge/setu/util"
 	"github.com/maticnetwork/heimdall/helper"
+	hmTypes "github.com/maticnetwork/heimdall/types"
 )
 
 // MaticChainListener - Listens to and process headerblocks from maticchain
 type MaticChainListener struct {
 	BaseListener
+	cacheLastSpan *hmTypes.Span
 }
 
 // NewMaticChainListener - constructor func
@@ -49,6 +51,12 @@ func (ml *MaticChainListener) Start() error {
 	// subscribed to new head
 	ml.Logger.Info("Subscribed to new head")
 
+	// cache last span
+	lastSpan, err := util.GetLastSpan(ml.cliCtx)
+	if err == nil && lastSpan != nil {
+		ml.cacheLastSpan = lastSpan
+	}
+
 	return nil
 }
 
@@ -70,10 +78,19 @@ func (ml *MaticChainListener) ProcessHeader(newHeader *types.Header) {
 }
 
 func (ml *MaticChainListener) checkAndSendSpanTask(newHeader *types.Header) {
+
+	// ignore header
+	if ml.cacheLastSpan != nil && newHeader.Number.Uint64() < ml.cacheLastSpan.StartBlock {
+		ml.Logger.Debug("Span already sent. ignore header", "blockNumber", newHeader.Number, "cacheSpanStartBlock", ml.cacheLastSpan.StartBlock)
+		return
+	}
+
 	// Fetch last span
 	lastSpan, err := util.GetLastSpan(ml.cliCtx)
 	if err == nil && lastSpan != nil {
 		ml.Logger.Debug("Found last span", "lastSpan", lastSpan.ID, "startBlock", lastSpan.StartBlock, "endBlock", lastSpan.EndBlock)
+		// update cache
+		ml.cacheLastSpan = lastSpan
 
 		// check if span task has to be sent
 		if lastSpan.StartBlock <= newHeader.Number.Uint64() && newHeader.Number.Uint64() <= lastSpan.EndBlock {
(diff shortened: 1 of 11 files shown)


bridge/setu/listener/maticchain.go
```diff
@@ -54,14 +55,42 @@ func (ml *MaticChainListener) Start() error {
 // ProcessHeader - process headerblock from maticchain
 func (ml *MaticChainListener) ProcessHeader(newHeader *types.Header) {
 	ml.Logger.Debug("New block detected", "blockNumber", newHeader.Number)
+
+	// check and send span task
+	go ml.checkAndSendSpanTask(newHeader)
```
jdkanani, line 60: This will fetch the last span very frequently. We should do span poll as we were doing before or optimize the last span fetch using cache.

bridge/setu/processor/checkpoint.go
```diff
@@ -292,7 +293,12 @@ func (cp *CheckpointProcessor) sendCheckpointAckToHeimdall(eventName string, che
 			return err
 		}
 	}
-	return nil
+	// After broadcasting transaction from bridge, add back the msg to queue with retry delay.
+	// This is to retry side-tx msg incase if it was failed earlier during side-tx processing on heimdall.
+	// estimatedNextBlockTime is in milliseconds
+	estimatedNextBlockTime := helper.GetGenesisDoc().ConsensusParams.Block.TimeIotaMs
+	cp.Logger.Debug("Retrying checkpoint-ack to check if side-tx is successful or not", "after", 6*time.Duration(estimatedNextBlockTime)*time.Millisecond)
+	return tasks.NewErrRetryTaskLater("retry to check if side-tx is successful or not", 6*time.Duration(estimatedNextBlockTime)*time.Millisecond)
```
jdkanani, line 301: This is very much the same during lifetime, so we can fetch on bridge start and use from memory.

bridge/setu/processor/clerk.go
```diff
@@ -112,7 +114,13 @@ func (cp *ClerkProcessor) sendStateSyncedToHeimdall(eventName string, logBytes s
...
 		}
 	}
-	return nil
+
+	// After broadcasting transaction from bridge, add back the msg to queue with retry delay.
+	// This is to retry side-tx msg incase if it was failed earlier during side-tx processing on heimdall.
+	// estimatedNextBlockTime is in milliseconds
+	estimatedNextBlockTime := helper.GetGenesisDoc().ConsensusParams.Block.TimeIotaMs
+	cp.Logger.Debug("Retrying deposit to check if side-tx is successful or not", "after", 6*time.Duration(estimatedNextBlockTime)*time.Millisecond)
+	return tasks.NewErrRetryTaskLater("retry to check if side-tx is successful or not", 6*time.Duration(estimatedNextBlockTime)*time.Millisecond)
```
jdkanani, line 123: This is very much the same during lifetime, so we can fetch on bridge start and use from memory.

bridge/setu/processor/span.go
```diff
@@ -1,109 +1,77 @@
...
-	if err != nil {
-		sp.Logger.Error("Unable to fetch current block", "error", err)
-		return
 	}
 
-	if lastSpan.StartBlock <= currentBlock && currentBlock <= lastSpan.EndBlock {
-		// log new span
-		sp.Logger.Info("✅ Proposing new span", "spanId", nextSpanMsg.ID, "startBlock", nextSpanMsg.StartBlock, "endBlock", nextSpanMsg.EndBlock)
+	// check and propose span
+	if lastSpan.StartBlock <= header.Number.Uint64() && header.Number.Uint64() <= lastSpan.EndBlock {
```
jdkanani, line 52: We are doing this twice now!

bridge/setu/processor/clerk.go
jdkanani: We can store in Bridge object.

bridge/setu/listener/maticchain.go
mankenavenkatesh: Added cache

bridge/setu/processor/span.go
mankenavenkatesh: This is required due to delay.

bridge/setu/listener/maticchain.go
```diff
@@ -48,20 +51,63 @@ func (ml *MaticChainListener) Start() error {
 	// subscribed to new head
 	ml.Logger.Info("Subscribed to new head")
 
+	// cache last span
+	lastSpan, err := util.GetLastSpan(ml.cliCtx)
+	if err == nil && lastSpan != nil {
+		ml.cacheLastSpan = lastSpan
```
jdkanani, line 57: `cacheLastSpan` never gets cleared?

PR #392, opened 2020-06-23
change delay interval

mankenavenkatesh:


bridge/setu/util/common.go
```diff
@@ -54,8 +54,8 @@ const (
 
 	TransactionTimeout      = 1 * time.Minute
 	CommitTimeout           = 2 * time.Minute
-	TaskDelayBetweenEachVal = 6 * time.Second
-	ValidatorJoinRetryDelay = 12 * time.Second
+	TaskDelayBetweenEachVal = 18 * time.Second
```
jdkanani, line 57: Can we have `BlockInterval = 6 * time.Second` and then use `TaskDelayBetweenEachVal = 3 * BlockInterval`?

PR #425, opened 2020-12-03
Removes unwanted modules `x/auth` and `x/blog`

### LINKED ISSUE #423: Remove dead code and examples
- Remove `x/auth` since we are using `cosmos-sdk/x/auth` and `x/blog` in favour of #422 
- Remove unwanted proto files

jdkanani:
Closes #423


app/app.go
```diff
@@ -159,7 +158,7 @@ func NewHeimdallApp(
 		paramstypes.StoreKey,
 	)
 	tkeys := sdk.NewTransientStoreKeys(paramstypes.TStoreKey)
-	memKeys := sdk.NewMemoryStoreKeys(blogtypes.MemStoreKey)
+	memKeys := sdk.NewMemoryStoreKeys()
```
anilCSE, line 161: are we using it anywhere?

app/app.go
```diff
@@ -169,7 +168,7 @@ func NewHeimdallApp(
 		invCheckPeriod:    invCheckPeriod,
 		keys:              keys,
 		tkeys:             tkeys,
-		memKeys:           memKeys,
+		// memKeys:           memKeys,
```
anilCSE, line 171: I would suggest to add a TODO if we gonna use them in future. If we are planning to enable IBC, we would need this for capability module
```suggestion
		// memKeys:           memKeys,
```

types/keys.go
```diff
@@ -12,4 +12,7 @@ const (
 var (
 	// CoinDecimals is the amount of staking tokens required for 1 unit
 	CoinDecimals = new(big.Int).Exp(big.NewInt(10), big.NewInt(18), nil)
+
+	// FeeToken is fee token symbol
+	FeeToken = "matic"
```
anilCSE, line 17: May be it shouldn't sit here.  How about making it configurable in genesis file directly? Can be addressed in a separate PR though

anilCSE (changes requested):
Loved this PR. Always :+1: for removing code, leff a few comments

types/keys.go
jdkanani: Agree. I will add it genesis as params. Which module should have it?

app/app.go
jdkanani: I will remove `memKeys` through out the code. We don't need it.

types/keys.go
anilCSE: It's so generic. I would suggest to keep it in auth/bank module. wdyt?

app/app.go
jdkanani: @anilCSE Fixed and pushed. Please review.

types/keys.go
jdkanani: Yeah we can have it in auth module. Can we add it in params even if we are using `cosmos/x/auth`?

jdkanani:
@anilCSE Please review.

types/keys.go
anilCSE: I think we cannot directly use `cosmos/x/auth`, it will fail in Params validation and Genesis Validation. I guess we can override them but that would be risky. Given this, `bank` looks a better option for our case.

anilCSE (approved):
lgtm :tada:

types/keys.go
jdkanani: Let's put that in the next PR. We can review/merge this PR in that case.

PR #434, opened 2020-12-07
chg: update golang version for ci

jdkanani:


.github/workflows/ci.yml
```diff
@@ -9,7 +9,9 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v1
```
anilCSE, line 10: lets use `actions/setup-go@v2.1.3`
```suggestion
        uses: actions/setup-go@v2.1.3
```

.github/workflows/ci.yml
```diff
@@ -9,7 +9,9 @@ jobs:
...
         with:
-          go-version: 1.14.7
+          go-version: 1.15.6
+      - name: Create golang symlink
+        run: sudo ln -f -s $GOROOT/bin/* /usr/bin/
```
anilCSE, line 14: why do we need this? I would suggest to avoid `sudo` unless necessary

jdkanani: It's because of this issue with Github actions - https://github.com/actions/virtual-environments/issues/731#issuecomment-617039721

jdkanani: New setup-go version fixed it. Removed sudo.

anilCSE (approved):

PR #448, opened 2020-12-09
Testcases for SideTx processor

jdkanani:


types/side_router.go
```diff
@@ -53,9 +53,6 @@ func (rtr *router) AddRoute(path string, h *SideHandlers) SideRouter {
 		panic("router sealed; cannot add route handler")
 	}
 
-	if !isAlphaNumeric(path) {
-		panic("route expressions can only contain alphanumeric characters")
-	}
```
anilCSE, line 58: don't we need this anymore? If so, please cleanup the unused declaration `isAlphaNumeric`

anilCSE (approved):
lgtm

types/side_router.go
jdkanani: Done.

PR #453, opened 2020-12-10
Add testnet command and clean cmd/ package for lint

jdkanani:


anilCSE:
@jdkanani imho, we don't need testnet command. It would be unnecessary maintainance burden. wdyt?

jdkanani:
Yeah, you are right. Closing this PR. I will reopen if we need it in the future.

PR #468, opened 2020-12-15
SideTxHandlers Integration

murthyvitwit:


x/staking/keeper/keeper.go
```diff
@@ -29,9 +29,9 @@ var (
 // ModuleCommunicator manages different module interaction
 type ModuleCommunicator interface {
 	GetACKCount(ctx sdk.Context) uint64
-	SetCoins(ctx sdk.Context, addr hmCommon.HeimdallAddress, amt sdk.Coins) error
-	GetCoins(ctx sdk.Context, addr hmCommon.HeimdallAddress) sdk.Coins
-	SendCoins(ctx sdk.Context, from hmCommon.HeimdallAddress, to hmCommon.HeimdallAddress, amt sdk.Coins) error
+	SetCoins(ctx sdk.Context, addr string, amt sdk.Coins) error
```
anilCSE, line 32: It should still be `sdk.AccAddress`.

anilCSE: Only proto messages and state will have them as strings.

x/staking/keeper/keeper.go
```diff
@@ -61,8 +61,8 @@ func NewKeeper(
 		cdc:                cdc,
 		storeKey:           storeKey,
 		paramSubspace:      paramstore,
-		chainKeeper:        chainKeeper,
-		moduleCommunicator: moduleCommunicator,
+		ChainKeeper:        chainKeeper,
+		ModuleCommunicator: moduleCommunicator,
```
anilCSE, line 65: I am not really comfortable with exporting `ModuleCommunicator`. @jdkanani ?

anilCSE (changes requested):

x/staking/keeper/keeper.go
jdkanani: That's tricky since side_handler is not in keeper package. Either we can keep it in keeper to avoid exporting moduleCommunicator, similar to normal handlers?

jdkanani (changes requested):

x/staking/keeper/keeper.go
```diff
@@ -29,9 +29,9 @@ var (
 // ModuleCommunicator manages different module interaction
 type ModuleCommunicator interface {
 	GetACKCount(ctx sdk.Context) uint64
-	SetCoins(ctx sdk.Context, addr hmCommon.HeimdallAddress, amt sdk.Coins) error
-	GetCoins(ctx sdk.Context, addr hmCommon.HeimdallAddress) sdk.Coins
-	SendCoins(ctx sdk.Context, from hmCommon.HeimdallAddress, to hmCommon.HeimdallAddress, amt sdk.Coins) error
+	SetCoins(ctx sdk.Context, addr string, amt sdk.Coins) error
+	GetCoins(ctx sdk.Context, addr string) sdk.Coins
```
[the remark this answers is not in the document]
anilCSE: Yes. There are risks with having direct access to entire Keeper, without proper authorization and ocaps. But hopefully SDK fixes this in coming release and we can update again after that

x/staking/keeper/keeper.go
anilCSE: normal handlers are also in different package (in <module> package). But let's not block on this, I will create a tracking issue and will see if we can find a better solution.

anilCSE: https://github.com/maticnetwork/heimdall/issues/469

x/staking/keeper/keeper.go
murthyvitwit: We need `Validator.Signer ` type for which `string` for above mothods. We have defined `Validator.PubKey` and `Validator.Signer` both as strings in `validator.proto`. Shoudl I change that?

anilCSE: We'll revisit this `string` notation. I am also not so comfortable using this. I found an hacky method, will test that out.

jdkanani (approved):

anilCSE (approved):

PR #471, opened 2020-12-17
Serve swagger docs

anilCSE:
- Rename `querier.proto` to `query.proto`
- Integrate grpc-gateway endpoint `/swagger/` in app.go


ssandeep (approved):

.gitignore
```diff
@@ -13,3 +13,6 @@ logs
 .idea
 .vscode
 .cache
+
+# swagger-gen
+/tmp-swagger-gen
```
jdkanani, line 18: Please add a new line at end of the file

client/docs/config.json
```diff
@@ -0,0 +1,43 @@
...
+            "url": "./client/docs/cosmos/params/v1beta1/query.swagger.json"
+        },
+        {
+            "url": "./client/docs/cosmos-legacy-swagger.yaml",
+            "dereference": {
+                "circular": "ignore"
+            }
+        }
+    ]
+}
```
jdkanani, line 43: Same here. Please new line at end of the file.

jdkanani (changes requested):

client/docs/cosmos-legacy-swagger.yaml
```diff
@@ -0,0 +1,2598 @@
+---
```
jdkanani, line 1: Do we need legacy file?

anilCSE: There are a few tendermint RPC endpoints, for which we don't have gRPC alertnatives. This legacy swagger has those endpoints. May be we can remove some APIs from there, will update

jdkanani: Got it.

jdkanani (approved):

PR #490, opened 2020-12-29
fix init chain

temaniarpit27:


proto/clerk/v1beta/clerk.proto
```diff
@@ -23,8 +24,6 @@ message EventRecord {
         (gogoproto.moretags) = "yaml:\"record_time\""
     ];
     uint64 log_index = 5;
-    string tx_hash   = 6
-        [(gogoproto.casttype) =
-             "github.com/maticnetwork/heimdall/types/common.HeimdallHash"];
-    string chain_id = 7;
+    string tx_hash   = 6;
```
anilCSE, line 27: Oh missed this, should we use bytes for hash? I believe we agreed to use string only for addresses
```suggestion
   bytes tx_hash   = 6
        [(gogoproto.casttype) =
             "github.com/maticnetwork/heimdall/types/common.HeimdallHash"];
```

anilCSE (commented):
lgtm, just one question

temaniarpit27:
@anilCSE : fixed review comment

proto/clerk/v1beta/clerk.proto
murthyvitwit: @anilCSE if we use `bytes` or `string` with `gogoproto.casttype` its throwing following error on `make hiemdall` command

`# github.com/maticnetwork/heimdall/x/clerk/types
x/clerk/types/clerk.pb.go:142:7: second argument to copy should be slice or string; have [32]byte
x/clerk/types/clerk.pb.go:427:13: cannot use append(m.TxHash[:0], dAtA[iNdEx:postIndex]...) (type []byte) as type "github.com/maticnetwork/heimdall/types/common".HeimdallHash in assignment
x/clerk/types/clerk.pb.go:428:16: invalid operation: m.TxHash == nil (mismatched types "github.com/maticnetwork/heimdall/types/common".HeimdallHash and nil)
x/clerk/types/clerk.pb.go:429:14: cannot use []byte literal (type []byte) as type "github.com/maticnetwork/heimdall/types/common".HeimdallHash in assignment
make: *** [make/release.mk:14: heimdall] Error 2
`

But It's strange that its not throwing **error** when I use it on `staking` module's `proto` types.

murthyvitwit (approved):

proto/clerk/v1beta/clerk.proto
anilCSE: Ah, we should fix this issue then. Let's have a follow-up PR

murthyvitwit: Sure

Issue #509, opened 2021-01-18
Review proto versioning


anilCSE:
- We are using `v1beta1` in upstream as we couldn't decide on stability of each module. We can rethink on this and if we feel like any module is stable from our end, we can remove the beta tag
- Also we are using `v1beta`, I believe it's more of a typo. We need to change them back to `v1beta1` if we don't have strong reasoing to change this
- Move proto files to `proto/heimdall` directory, ex: bank proto files sits in `proto/heimdall/bank`

anilCSE:
cc @ssandeep @murthyvitwit

PR #554, opened 2021-02-15
Proto versioning 

saikumar0x1967:
#509


ssandeep (changes requested):
We will have to change the swagger config file also. Any other places we need to make changes?

saikumar0x1967:
@ssandeep  i changed proto version in swagger config also

anilCSE (approved):
:tada:
thanks for this pr @saikumar0x1967

ssandeep (approved):

Issue #581, opened 2021-03-03
Migrate CheckpointACKTxCmd


ssandeep:
It is currently commented out (checkpoint/client/cli/tx.go). Needs to be migrated and tested.

PR #582, opened 2021-03-03
added CheckpointACKTxCmd to tx cli of checkpoint

saikumar0x1967:
#581


ssandeep (approved):
Any additional test cases we need to add/uncomment for this?

saikumar0x1967:
@ssandeep  no need additional test cases for this change

PR #612, opened 2021-03-17
reduce the heimdall container size

saikumar0x1967:


ssandeep (approved):

jdkanani:
@saikumar0x1967 container shouldn't have `start` command. Because it will prevent other command in docker image.

PR #715, opened 2021-07-15
Bridge nonce issue fix

0xKrishna:


bridge/setu/processor/staking.go
```diff
@@ -333,3 +385,61 @@ func (sp *StakingProcessor) isOldTx(cliCtx cliContext.CLIContext, txHash string,
...
+
+	stakingTxnCount, err := queryTxCount(sp.cliCtx, validatorId, currentHeight)
+	if err != nil {
+		sp.Logger.Error("Failed to query stake txns by txquery for the given validator", "validatorId", validatorId)
+		return false, 0, err
+	}
+
+	if stakingTxnCount != 0 {
+		sp.Logger.Info("Recent staking txn count for the given validator is not zero", "validatorId", validatorId, "currentNonce", currentNonce, "txnNonce", txnNonce, "currentHeight", currentHeight)
+		return false, 1, nil
```
ssandeep, line 409: @0xKrishna , why are we returning 1 here, sorry if I forgot this from our last discussion.

ssandeep (approved):

bridge/setu/processor/staking.go
0xKrishna: To make the incoming wait for 1 unit of time (i.e. 10 Seconds).
Why 1 unit only? Because if the txn has passed the nonce check that means that it's the next txn in line and we should minimize the wait in which it will come back again.

PR #720, opened 2021-08-13
Max gas price check

0xKrishna:


helper/tx.go
```diff
@@ -72,8 +80,7 @@ func (c *ContractCaller) SendCheckpoint(signedData []byte, sigs [][3]*big.Int, r
 	auth, err := GenerateAuthObj(GetMainClient(), rootChainAddress, data)
 	if err != nil {
 		Logger.Error("Unable to create auth object", "error", err)
-		Logger.Info("Setting custom gaslimit", "gaslimit", GetConfig().MainchainGasLimit)
-		auth.GasLimit = GetConfig().MainchainGasLimit
```
ssandeep, line 76: What was this code doing? It's only the gas price that we are limiting through our config right, is there any change in the gas limit logic?

0xKrishna: In case of failure to create auth object, Earlier this code was creating one with MainchainGasLimit and ignoring the error, Hence even in case of error, the txn flow will go through. 

I had to remove it because we want the flow to return an error in case of high gas price (more than the limit), So I updated the code and returned the error from there instead of letting it go through.

ssandeep (approved):

PR #730, opened 2021-10-04
add checkpoint-adjust

temaniarpit27:


checkpoint/handler.go
```diff
@@ -30,6 +32,54 @@ func NewHandler(k Keeper, contractCaller helper.IContractCaller) sdk.Handler {
...
+// handleMsgCheckpointAdjust adjusts checkpoint
+func handleMsgCheckpointAdjust(ctx sdk.Context, msg types.MsgCheckpointAdjust, k Keeper, contractCaller helper.IContractCaller) sdk.Result {
+	logger := k.Logger(ctx)
+	chainParams := k.ck.GetParams(ctx).ChainParams
+	params := k.GetParams(ctx)
+
+	checkpointBuffer, err := k.GetCheckpointFromBuffer(ctx)
+	if checkpointBuffer != nil {
+		logger.Error("checkpoint buffer", "error", err)
+		return common.ErrNoCheckpointFound(k.Codespace()).Result()
```
jdkanani, line 44:
```suggestion
		logger.Error("checkpoint buffer exists", "error", err)
		return common.ErrCheckpointBufferFound(k.Codespace()).Result()
```

checkpoint/handler.go
```diff
@@ -30,6 +32,54 @@ func NewHandler(k Keeper, contractCaller helper.IContractCaller) sdk.Handler {
...
+
+	root, start, end, _, _, err := contractCaller.GetHeaderInfo(msg.HeaderIndex, rootChainInstance, params.ChildBlockInterval)
+	if err != nil {
+		logger.Error("Unable to fetch checkpoint from rootchain", "error", err, "checkpointNumber", msg.HeaderIndex)
+		return common.ErrNoCheckpointFound(k.Codespace()).Result()
+	}
+
+	if checkpointObj.EndBlock == end && checkpointObj.StartBlock == start && bytes.Equal(checkpointObj.RootHash.Bytes(), root.Bytes()) {
+		logger.Error("Same Checkpoint in DB")
+		return common.ErrOldCheckpoint(k.Codespace()).Result()
```
jdkanani, line 67:
```suggestion
		return common.ErrCheckpointAlreadyExists(k.Codespace()).Result()
```

checkpoint/side_handler.go
```diff
@@ -32,6 +34,45 @@ func NewSideTxHandler(k Keeper, contractCaller helper.IContractCaller) hmTypes.S
...
+	root, start, end, _, _, err := contractCaller.GetHeaderInfo(msg.HeaderIndex, rootChainInstance, params.ChildBlockInterval)
+	if err != nil {
+		logger.Error("Unable to fetch checkpoint from rootchain", "error", err, "checkpointNumber", msg.HeaderIndex)
+		return common.ErrorSideTx(k.Codespace(), common.CodeOldCheckpoint)
+	}
+
+	if checkpointObj.EndBlock == end && checkpointObj.StartBlock == start && bytes.Equal(checkpointObj.RootHash.Bytes(), root.Bytes()) {
+		logger.Error("Same Checkpoint in DB")
+		return common.ErrorSideTx(k.Codespace(), common.CodeOldCheckpoint)
+	}
```
jdkanani, line 70: We have to check if `msg` has the same data as `root, start, end, _, _, err` because we will be saving `msg` in the post handler `PostHandleMsgCheckpointAdjust`

checkpoint/side_handler.go
```diff
@@ -124,6 +167,78 @@ func NewPostTxHandler(k Keeper, contractCaller helper.IContractCaller) hmTypes.P
...
+	if err != nil {
+		logger.Error("Unable to fetch rootchain contract instance", "error", err)
+		return common.ErrNoCheckpointFound(k.Codespace()).Result()
+	}
+
+	root, start, end, _, _, err := contractCaller.GetHeaderInfo(msg.HeaderIndex, rootChainInstance, params.ChildBlockInterval)
+	if err != nil {
+		logger.Error("Unable to fetch checkpoint from rootchain", "error", err, "checkpointNumber", msg.HeaderIndex)
+		return common.ErrNoCheckpointFound(k.Codespace()).Result()
+	}
```
jdkanani, line 204: This call shouldn't be here. We cannot call external sources from PostHandler, only in SideHandler. So we will check if msg is the same as data on Ethereum chain in Side handler and in PostHandler we will validate data same as Handler and store checkpointObj, that's it. 

```suggestion
```

checkpoint/types/msg.go
```diff
@@ -18,6 +18,46 @@ import (
 
 var _ sdk.Msg = &MsgCheckpoint{}
 
+// MsgCheckpointAdjust represents checkpoint adjust
+type MsgCheckpointAdjust struct {
+	HeaderIndex uint64 `json:"header_index"`
+}
```
jdkanani, line 24: This has to be the same as `MsgCheckpointAck` because we need to use msg value to store it in db in `MsgPostHandlerCheckpointAdjust`

```suggestion
// MsgCheckpointAdjust represents checkpoint adjust
type MsgCheckpointAdjust struct {
	From       types.HeimdallAddress `json:"from"`
	Number     uint64                `json:"number"`
	Proposer   types.HeimdallAddress `json:"proposer"`
	StartBlock uint64                `json:"start_block"`
	EndBlock   uint64                `json:"end_block"`
	RootHash   types.HeimdallHash    `json:"root_hash"`
	TxHash     types.HeimdallHash    `json:"tx_hash"`
	LogIndex   uint64                `json:"log_index"`
}
```

jdkanani (changes requested):

checkpoint/handler.go
temaniarpit27: changing errors already

checkpoint/handler.go
temaniarpit27: changing errors already

jdkanani:
Also, let's add test cases for this since it's very hard to test in testnet

checkpoint/side_handler.go
```diff
@@ -32,6 +34,39 @@ func NewSideTxHandler(k Keeper, contractCaller helper.IContractCaller) hmTypes.S
...
+	root, start, end, _, proposer, err := contractCaller.GetHeaderInfo(msg.HeaderIndex, rootChainInstance, params.ChildBlockInterval)
+	if err != nil {
+		logger.Error("Unable to fetch checkpoint from rootchain", "error", err, "checkpointNumber", msg.HeaderIndex)
+		return common.ErrorSideTx(k.Codespace(), common.CodeNoCheckpoint)
+	}
+
+	if msg.EndBlock == end && msg.StartBlock == start && bytes.Equal(msg.RootHash.Bytes(), root.Bytes()) && bytes.Equal(msg.Proposer.Bytes(), proposer.Bytes()) {
+		logger.Error("Same Checkpoint in DB")
+		return common.ErrorSideTx(k.Codespace(), common.CodeCheckpointAlreadyExists)
+	}
```
jdkanani, line 64: We want this to be the same. 


```suggestion
	if msg.EndBlock != end || msg.StartBlock != start || !bytes.Equal(msg.RootHash.Bytes(), root.Bytes()) && bytes.Equal(msg.Proposer.Bytes(), proposer.Bytes()) {
		logger.Error("Checkpoint on Rootchain is not same as msg")
		return common.ErrorSideTx(k.Codespace(), common.CodeCheckpointAlreadyExists)
	}
```

jdkanani (changes requested):

checkpoint/side_handler.go
temaniarpit27: ha ye sahi kr dia maine
testing ke waqt dekha

jdkanani:
@temaniarpit27  Please add test cases in side_handler_test.go too

ssandeep (approved):

PR #738, opened 2021-11-04
POS-15 Remove building Heimdall for each chain

igorcrevar:
a new Heimdall binary gets built for each chain using Makefile and including some config parameters at build time.

Build is done with:

https://github.com/maticnetwork/heimdall#usage

Heimdall runs this function to generate the heimdall_params.go:

https://github.com/maticnetwork/heimdall/blob/master/Makefile#L34

https://github.com/maticnetwork/heimdall/blob/master/helper/heimdall-params.template.go

It is only used to set this param for the client:

https://github.com/maticnetwork/heimdall/blob/master/helper/heimdall-params.template.go#L20

The goal is to move those parameters to Heimdall and use a flag:

$ heimdall --chain=[mumbai,mainnet,local]

to run the binary for each chain. 

Internally, depending on the chain set the value of the NewSelectionAlgoHeight value.

Proposal:
There are multiple copying of chain toml files (during build and during init), maybe we should hard-code NewSelectionAlgoHeights and create files (or just one toml) during init phase


cmd/heimdallcli/main.go
```diff
@@ -60,11 +60,13 @@ func initTendermintViperConfig(cmd *cobra.Command) {
 	tendermintNode, _ := cmd.Flags().GetString(helper.NodeFlag)
 	homeValue, _ := cmd.Flags().GetString(helper.HomeFlag)
 	withHeimdallConfigValue, _ := cmd.Flags().GetString(helper.WithHeimdallConfigFlag)
+	networkChain, _ := cmd.Flags().GetString(helper.NetworkChainFlag)
 
 	// set to viper
 	viper.Set(helper.NodeFlag, tendermintNode)
 	viper.Set(helper.HomeFlag, homeValue)
 	viper.Set(helper.WithHeimdallConfigFlag, withHeimdallConfigValue)
+	viper.Set(helper.NetworkChainFlag, networkChain)
```
ferranbt, line 69: Same as before, where is this flag used?

cmd/heimdalld/init.go
```diff
@@ -50,6 +50,12 @@ func initCmd(ctx *server.Context, cdc *codec.Codec) *cobra.Command {
 
 			WriteDefaultHeimdallConfig(filepath.Join(config.RootDir, "config/heimdall-config.toml"), helper.GetDefaultHeimdallConfig())
 
+			// copy toml files for each network chain into config folder
+			err = helper.CopyNetworkChainTomlsToConfigDir(filepath.Join(config.RootDir, "config"), helper.GetValidNetworkChains())
```
ferranbt, line 54: what is the goal of this?

helper/config.go
```diff
@@ -107,6 +110,9 @@ type Configuration struct {
 
 	// wait time related options
 	NoACKWaitTime time.Duration `mapstructure:"no_ack_wait_time"` // Time ack service waits to clear buffer and elect new proposer
+
+	// current network chain
+	NetworkChain string `mapstructure:"network_chain"`
```
ferranbt, line 115: Can we call it "chain" instead? It is similar to what we use already in Bor.

ferranbt (changes requested):

helper/config.go
```diff
@@ -204,6 +213,14 @@ func InitHeimdallConfigWith(homeDir string, heimdallConfigFilePath string) {
 	privVal := privval.LoadFilePV(filepath.Join(configDir, "priv_validator_key.json"), filepath.Join(configDir, "priv_validator_key.json"))
 	cdc.MustUnmarshalBinaryBare(privVal.Key.PrivKey.Bytes(), &privObject)
 	cdc.MustUnmarshalBinaryBare(privObject.PubKey().Bytes(), &pubObject)
+
+	// get network chain form viper/cobra flag and set newSelectionAlgoHeight
+	networkChain := viper.GetString(NetworkChainFlag)
+	if networkChain != "" && isValidNetworkChain(networkChain) {
+		newSelectionAlgoHeight = readNetworkChainToml(configDir, networkChain)
```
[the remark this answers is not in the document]
igorcrevar: Should I put those numbers only in code or should they exist in some config file also?

cmd/heimdalld/init.go
igorcrevar: this copies mainnet.toml and mumbai.toml from build/app directory to config directory. If we hardcode things, this is not needed

victorcoder (approved):
LGTM

igorcrevar:
I forgot to edit README.md and to remove one line from .gitignore and makefile referencing heimdall-params.go
Also I have added $ heimdalld init between install and start part of readme, it was not there before. Please review these changes also

ferranbt:
LGTM! I will add also Arpit since he is the one with the most experience with Heimdall.

ferranbt (approved):

ssandeep (approved):

ferranbt:
As commented over Slack, please change the target branch for the PR to v2 and we can merge it.

PR #742, opened 2021-11-09
POS-24 Embed Heimdall bridge in the Daemon binary

igorcrevar:
Right now, there are three Heimdall services that we have to run in the network:
- Daemon
- Bridge
- Rest server 

Rest server is defined in the Daemon binary and runs as:
```bash
$ heimdalld rest-server
```
However, the Bridge is an independent binary.
This makes it complex to distribute the software (release binaries and docker images).

The goal of this issue is to integrate the bridge on the Daemon alongside the Rest server in a command like this:
```bash
$ heimdalld bridge
```

### DIFF
--- a/bridge/cmd/root.go
+++ b/bridge/cmd/root.go
@@ -32,7 +32,7 @@ func BridgeCommands() *cobra.Command {
 
 // initTendermintViperConfig sets global viper configuration needed to heimdall
 func initTendermintViperConfig(cmd *cobra.Command) {
-	tendermintNode, _ := cmd.Flags().GetString(helper.NodeFlag)
+	heimdallNode, _ := cmd.Flags().GetString(helper.HeimdallNodeFlag)
 	homeValue, _ := cmd.Flags().GetString(helper.HomeFlag)
 	withHeimdallConfigValue, _ := cmd.Flags().GetString(helper.WithHeimdallConfigFlag)
 	bridgeDBValue, _ := cmd.Flags().GetString(bridgeDBFlag)
@@ -44,7 +44,7 @@ func initTendermintViperConfig(cmd *cobra.Command) {
 	}
 
 	// set to viper
-	viper.Set(helper.NodeFlag, tendermintNode)
+	viper.Set(helper.HeimdallNodeFlag, heimdallNode)
 	viper.Set(helper.HomeFlag, homeValue)
 	viper.Set(helper.WithHeimdallConfigFlag, withHeimdallConfigValue)
 	viper.Set(bridgeDBFlag, bridgeDBValue)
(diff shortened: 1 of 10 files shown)


igorcrevar:
In root.go of rest server the default value for listen address flag is defined as tcp protocol:
```
cmd.Flags().String(client.FlagListenAddr, "tcp://0.0.0.0:1317", "The address for the server to listen on")
```
Should it be http (defined in config.go in helper package as ```DefaultHeimdallServerURL = "http://0.0.0.0:1317"```) ? 

Should we add bridge and rest-server commands into readme and how?

bridge/cmd/root.go
```diff
@@ -50,19 +54,10 @@ func InitTendermintViperConfig(cmd *cobra.Command) {
...
-		fmt.Println(err)
-		os.Exit(1)
-	}
-}
-
 func init() {
 	var logger = helper.Logger.With("module", "bridge/cmd/")
-	rootCmd.PersistentFlags().StringP(helper.NodeFlag, "n", "tcp://localhost:26657", "Node to connect to")
-	rootCmd.PersistentFlags().String(helper.HomeFlag, os.ExpandEnv("$HOME/.heimdalld"), "directory for config and data")
+	rootCmd.PersistentFlags().StringP(helper.NodeFlag, "n", helper.DefaultNode, "Node to connect to")
```
ferranbt, line 59: Could we have a more descriptive name of what is this? Is this the connection to the Heimdall daemon? @temaniarpit27 @0xKrishna

ferranbt:
LGTM! For the README just add another shell command after the rest part. I will ping @temaniarpit27 for the tcp issue since I am not sure what it means. Also, please check the other comment in the code.

bridge/cmd/root.go
temaniarpit27: 26657 is the tendermint port

victorcoder: `Tendermint` looks too specific, what about `HeimdallNode`?

victorcoder (approved):
LGTM

bridge/cmd/root.go
igorcrevar: I could change then to ```HeimdallNodeFlag```,  ```DefaultHeimdallNode``` . There will be changes in 7 files because of renaming. Is it ok to keep in server/root.go
```cmd.Flags().String(client.FlagNode, helper.DefaultHeimdallNode, "Address of the node to connect to")```
instead of 
```cmd.Flags().String(helper.HeimdallNodeFlag, helper.DefaultHeimdallNode, "Address of the node to connect to")``` because all other flags names are retrieved from tinderment client package?

igorcrevar: I have changed name of the flag and default value const to Tendermint and pushed.
I'll mark this as resolved
If I am correct with wrong default value for FlagListenAddr flag we can change it through this or some other task

temaniarpit27 (approved):
LGTM

ferranbt (approved):

PR #743, opened 2021-11-11
Heimdall crashes with "too many open files"

temaniarpit27:


temaniarpit27:
I dont think this will make a diff as we are not facing the issue in bridge but in heimdalld but still I have made the change

victorcoder (approved):
LGTM

temaniarpit27:
> HTTP client leak, good catch! LGTM

Testing it on a node to see if that's really the fix or not

PR #746, opened 2021-11-15
Allow Heimdall to be configured using env vars

victorcoder:
Thanks to viper we can setup the automatic env configuration, this is very convenient for running in docker as we don't need to add explicit params to the CLI and we avoid having to sed on config files.

### DIFF
--- a/helper/config.go
+++ b/helper/config.go
@@ -163,6 +163,8 @@ func InitHeimdallConfigWith(homeDir string, heimdallConfigFilePath string) {
 	configDir := filepath.Join(homeDir, "config")
 
 	heimdallViper := viper.New()
+	heimdallViper.SetEnvPrefix("HEIMDALL")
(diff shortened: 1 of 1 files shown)


ferranbt:
Looks good. Though we are still limited because we do not have flags for many of the fields, only the one that points to the config file.

victorcoder:
> Looks good. Though we are still limited because we do not have flags for many of the fields, only the one that points to the config file.

@ferranbt yes, but I'm currently looking to try to start up using docker/compose, without touching any config file at startup time, making the UX as simple as `docker-compose up` while we maintain rabbitmq and without compose when we get rid of it.

So for the first goal it works.

ferranbt:
> So for the first goal it works.

Cool!

ferranbt (approved):

victorcoder:
Checked and working well

PR #748, opened 2021-11-22
Refactor CI workflows to publish Heimdall

vcastellm:
Create release, packages, create and push docker images using Goreleaser workflow.

## Implementation

As we are going to embed genesis data into the binary, this is the most similar method to do it before the new CLI release.

This way, each release gets tied to its corresponding genesis file.

This establish `heimdall` repository as the source of truth for the genesis file.


Makefile
```diff
@@ -131,3 +131,41 @@ build-docker-develop:
 	docker build -t "maticnetwork/heimdall:develop" -f docker/Dockerfile.develop .
 
 .PHONY: contracts build
+
+PACKAGE_NAME          := github.com/maticnetwork/heimdall
+GOLANG_CROSS_VERSION  ?= v1.17.3
```
ssandeep, line 136: Any issues if a server is using a different minor version of go?

ssandeep (approved):

Makefile
vcastellm: This only applies to the CI used to build the bins, the idea is to stop installing Go and building the binaries on the servers to move towards immutability.

.goreleaser.yml
```diff
@@ -0,0 +1,270 @@
...
+      - heimdallcli-linux-arm64
+      - bridge-linux-arm64
+    build_flag_templates:
+      - --platform=linux/arm64
+    skip_push: false
+    extra_files:
+      - docker/entrypoint.sh
+
+docker_manifests:
+  - name_template: 0xpolygon/{{ .ProjectName }}:{{ .Version }}
```
ferranbt, line 256: Are we publishing in 0xPolygon docker org?

vcastellm: That's the goal after the announcement and bump to v0.3.x, yes

PR #749, opened 2021-11-25
POS-50: Run Rest and Bridge service alongside the Heimdall daemon.

0xSasaPrsic:
- added two flags --rest-server and --bridge for starting services alongside Heimdall daemon
- heimdall start command moved from Tendermint SDK to Heimdall in order to be able to modify it more freely
- added rest server and bridge async start ordering logic using channels
- added health check for rest server using open port check (for service ordering purposes)
- added default log level for bridge in order to bypass a Tendermint SDK function panic when no log level has been passed from flags and level is read from config.toml

Order of services graceful shutting down is yet to come.

### DIFF
--- a/cmd/heimdalld/main.go
+++ b/cmd/heimdalld/main.go
@@ -8,27 +8,41 @@ import (
 	"math/big"
 	"os"
 	"path/filepath"
+	"runtime/pprof"
 	"strings"
 	"time"
 
 	"github.com/cosmos/cosmos-sdk/baseapp"
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
 	"github.com/cosmos/cosmos-sdk/codec"
 	"github.com/cosmos/cosmos-sdk/server"
 	"github.com/cosmos/cosmos-sdk/store"
+
+	bridgeCmd "github.com/maticnetwork/heimdall/bridge/cmd"
+	restServer "github.com/maticnetwork/heimdall/server"
+	tserver "github.com/tendermint/tendermint/abci/server"
+
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	ethCommon "github.com/maticnetwork/bor/common"
 	hmbridge "github.com/maticnetwork/heimdall/bridge/cmd"
+	"github.com/maticnetwork/heimdall/version"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	abci "github.com/tendermint/tendermint/abci/types"
+	tcmd "github.com/tendermint/tendermint/cmd/tendermint/commands"
+
 	cfg "github.com/tendermint/tendermint/config"
 	"github.com/tendermint/tendermint/crypto"
 	"github.com/tendermint/tendermint/crypto/secp256k1"
 	"github.com/tendermint/tendermint/libs/cli"
 	"github.com/tendermint/tendermint/libs/common"
 	"github.com/tendermint/tendermint/libs/log"
+	"github.com/tendermint/tendermint/node"
 	"github.com/tendermint/tendermint/p2p"
 	"github.com/tendermint/tendermint/privval"
+	pvm "github.com/tendermint/tendermint/privval"
+	"github.com/tendermint/tendermint/proxy"
 	tmTypes "github.com/tendermint/tendermint/types"
 	tmtime "github.com/tendermint/tendermint/types/time"
 	dbm "github.com/tendermint/tm-db"
@@ -53,6 +67,17 @@ var (
 	flagNodeHostPrefix   = "node-host-prefix"
 )
 
+// Tendermint full-node start flags
+const (
+	flagWithTendermint = "with-tendermint"
+	flagAddress        = "address"
+	flagTraceStore     = "trace-store"
+	flagPruning        = "pruning"
+	flagCPUProfile     = "cpu-profile"
+	FlagMinGasPrices   = "minimum-gas-prices"
+	FlagHaltHeight     = "halt-height"
+	FlagHaltTime       = "halt-time"
+)
 const (
 	nodeDirPerm = 0755
 )
@@ -87,6 +112,11 @@ func main() {
 		PersistentPreRunE: server.PersistentPreRunEFn(ctx),
 	}
 
+	tendermintCmd := &cobra.Command{
+		Use:   "tendermint",
+		Short: "Tendermint subcommands",
+	}
+
 	// add new persistent flag for heimdall-config
 	rootCmd.PersistentFlags().String(
 		helper.WithHeimdallConfigFlag,
@@ -108,7 +138,23 @@ func main() {
 		logger.Error("main | BindPFlag | helper.ChainFlag", "Error", err)
 	}
 
-	server.AddCommands(ctx, cdc, rootCmd, newApp, exportAppStateAndTMValidators)
+	rootCmd.AddCommand(heimdallStart(ctx, newApp, cdc)) // New Heimdall start command
+
+	tendermintCmd.AddCommand(
+		server.ShowNodeIDCmd(ctx),
+		server.ShowValidatorCmd(ctx),
+		server.ShowAddressCmd(ctx),
+		server.VersionCmd(ctx),
+	)
+
+	rootCmd.AddCommand(server.UnsafeResetAllCmd(ctx))
+	rootCmd.AddCommand(flags.LineBreak)
+	rootCmd.AddCommand(tendermintCmd)
+	rootCmd.AddCommand(server.ExportCmd(ctx, cdc, exportAppStateAndTMValidators))
+	rootCmd.AddCommand(flags.LineBreak)
+	rootCmd.AddCommand(version.Cmd) // Using heimdall version, not Cosmos SDK version
+	// End of block
+
 	rootCmd.AddCommand(showAccountCmd())
 	rootCmd.AddCommand(showPrivateKeyCmd())
 	rootCmd.AddCommand(hmserver.ServeCommands(cdc, hmserver.RegisterRoutes))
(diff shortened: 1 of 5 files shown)


server/root.go
```diff
@@ -22,6 +24,24 @@ import (
 	_ "github.com/maticnetwork/heimdall/server/statik"
 )
 
+func StartRestServer(cdc *codec.Codec, registerRoutesFn func(*lcd.RestServer), restCh chan struct{}) error {
+	rs := lcd.NewRestServer(cdc)
+	registerRoutesFn(rs)
+	logger := tmLog.NewTMLogger(log.NewSyncWriter(os.Stdout)).With("module", "rest-server")
+	go restServerHealthCheck(restCh)
```
ferranbt, line 31: What happens if the daemon is not running? Does it crashes? I do not think so because I have seen heimdall stopping before and the rest server still running (I think).

cmd/heimdalld/main.go
```diff
@@ -138,6 +184,254 @@ func exportAppStateAndTMValidators(logger log.Logger, db dbm.DB, storeTracer io.
...
+			pprof.StopCPUProfile()
+			f.Close()
+		}
+	}
+
+	// start rest
+	if startRestServer {
+		restCh := make(chan struct{})
+		go func() {
+			_ = restServer.StartRestServer(cdc, hmserver.RegisterRoutes, restCh)
```
ferranbt, line 387: Could this just be instead something like this:
```
closeCh := make(chan struct{})
go func() {
    _ = restServer.StartRestServer(cdc, hmserver.RegisterRoutes)
    close(closeCh)
}()
<-closeCh
```
I think this change would not affect either the other call to `StartRestServer`.

cmd/heimdalld/main.go
```diff
@@ -138,6 +184,254 @@ func exportAppStateAndTMValidators(logger log.Logger, db dbm.DB, storeTracer io.
...
+	}
+	// bridge flags
+	cmd.Flags().Bool("all", false, "start all bridge services")
+	cmd.Flags().StringSlice("only", []string{}, "comma separated bridge services to start")
+
+	// rest server flags
+	cmd.Flags().String(client.FlagListenAddr, "tcp://0.0.0.0:1317", "The address for the server to listen on")
+	cmd.Flags().Bool(client.FlagTrustNode, true, "Trust connected full node (don't verify proofs for responses)")
+	cmd.Flags().String(client.FlagChainID, "", "The chain ID to connect to")
+	cmd.Flags().String(client.FlagNode, helper.DefaultTendermintNode, "Address of the node to connect to")
```
ferranbt, line 260: Is this needed? It should already know the target address for the Tendermint node in the code, I think it is one of the flags to bind.

cmd/heimdalld/main.go
```diff
@@ -138,6 +184,254 @@ func exportAppStateAndTMValidators(logger log.Logger, db dbm.DB, storeTracer io.
...
+		logger.Error("main | BindPFlag | helper.LogLevel", "Error", err)
+	}
+	// bridge flags
+	cmd.Flags().Bool("all", false, "start all bridge services")
+	cmd.Flags().StringSlice("only", []string{}, "comma separated bridge services to start")
+
+	// rest server flags
+	cmd.Flags().String(client.FlagListenAddr, "tcp://0.0.0.0:1317", "The address for the server to listen on")
+	cmd.Flags().Bool(client.FlagTrustNode, true, "Trust connected full node (don't verify proofs for responses)")
+	cmd.Flags().String(client.FlagChainID, "", "The chain ID to connect to")
```
ferranbt, line 259: What is the usage of this flag? Is this rest specific or Heimdall?

cmd/heimdalld/main.go
```diff
@@ -138,6 +184,254 @@ func exportAppStateAndTMValidators(logger log.Logger, db dbm.DB, storeTracer io.
...
+	cmd.Flags().Uint64(FlagHaltHeight, 0, "Height at which to gracefully halt the chain and shutdown the node")
+	cmd.Flags().Uint64(FlagHaltTime, 0, "Minimum block time (in Unix seconds) at which to gracefully halt the chain and shutdown the node")
+	cmd.Flags().String(flagCPUProfile, "", "Enable CPU profiling and write to the provided file")
+
+	// add support for all Tendermint-specific command line options
+	tcmd.AddNodeFlags(cmd)
+	return cmd
+}
+
+func startStandAlone(ctx *server.Context, appCreator server.AppCreator) error {
```
ferranbt, line 281: What is the difference between `startStandAlone` and `startInProcess`?

cmd/heimdalld/main.go
```diff
@@ -138,6 +184,254 @@ func exportAppStateAndTMValidators(logger log.Logger, db dbm.DB, storeTracer io.
...
+		}
+
+		if cpuProfileCleanup != nil {
+			cpuProfileCleanup()
+		}
+
+		ctx.Logger.Info("exiting...")
+	})
+
+	// run forever (the node will not be returned)
```
ferranbt, line 413: Is this the thing you mentioned about graceful shutdown right? Please add a TODO.

ferranbt (changes requested):

server/root.go
sh3ll3x3c: Rest server doesn't crash without Heimdall daemon, but bridge panics if rest server hasn't been started, that's why we put in the health check.

cmd/heimdalld/main.go
sh3ll3x3c: startRestServer function is blocking, so the channel would never be executed, in the example you provided...

cmd/heimdalld/main.go
sh3ll3x3c: It's a Heimdall flag, we put it under rest by mistake...

cmd/heimdalld/main.go
sprsic: This was moved from cosmos SDK, and it will start ABCI with or without Tendermint depending on the flag **with-tendermint**.

cmd/heimdalld/main.go
sprsic: yes, I will add, this function is blocking, and we will have to change the logic in server.TrapSignal so that there is no os.Exit since it will prevent gracefully shutdown of the rest server, bridge.

server/root.go
ferranbt: I see now, you use rest as a done channel to notify when rest server is running.

cmd/heimdalld/main.go
ferranbt: This is a copy paste from cosmos right? You did not change anything.

sh3ll3x3c: Yeah, but we're just using the startInProcess func, which means that we can potentially delete this one without any consequences...

cmd/heimdalld/main.go
0xSasaPrsic: yes, I will add, this function is blocking, and we will have to change the logic in server.TrapSignal so that there is no os.Exit since it will prevent gracefully shutdown of the rest server, bridge.

cmd/heimdalld/main.go
0xSasaPrsic: This flag was defined for rest-server in cosmos sdk, so it was copied here. Agree, this is for the flag to bind and if not provided it will use default one.

vcastellm (approved):
LGTM but there are some old thing that I don't get, for example https://github.com/maticnetwork/heimdall/blob/d20bc3909567f1e12c0b121091f43fd257209b92/bridge/cmd/start.go#L131 is the `--all` flag necessary for the bridge? how this will be reflected in the new `--bridge` flag in main command?

cmd/heimdalld/main.go
ferranbt: Ok, lets delete it then.

0xSasaPrsic:
> LGTM but there are some old thing that I don't get, for example
> 
> https://github.com/maticnetwork/heimdall/blob/d20bc3909567f1e12c0b121091f43fd257209b92/bridge/cmd/start.go#L131
> is the `--all` flag necessary for the bridge? how this will be reflected in the new `--bridge` flag in main command?

hey @vcastellm when starting a bridge with legacy command for starting a bridge service, it is necessary to pass a flag `--all` or `--only` - with csv values of the processors this was not changed. In order to start bridge with new flag added to` heimdalld start` command, it is also possible to pass `--all` or `--only` flags in order for the bridge to start. If we don't want this flexibility we could hardcode the flag values.

ferranbt (approved):

vcastellm:
> If we don't want this flexibility we could hardcode the flag values.

Is it useful to use --only?

0xSasaPrsic:
> > If we don't want this flexibility we could hardcode the flag values.
> 
> Is it useful to use --only?
With `--only` we would have possibility to include only subset of 6 available processors: checkpointProcessor, stakingProcessor, clerkProcessor, feeProcessor, spanProcessor, slashingProcessor. If we don't want this, we could remove `--only` and  hardcode `--all` value which will include all 6 processors.

vcastellm:
Oh sorry, I mean just for the bridge, only two components no?

0xSasaPrsic:
> Oh sorry, I mean just for the bridge, only two components no?

I am not exactly sure about the last question.
Currently the implementation of starting Bridge within the heimdall start command goes like this:
When --bridge flag is submitted one of the two following flags MUST be submitted as well: --all or --only. These are legacy flags for Bridge, otherwise there will be panic during the Bridge startup.
Maybe we can change this behavior in the way that if there is no --all or --only flag we assume that --all flag is a default value?

vcastellm:
I mean, bridge only has two services https://github.com/maticnetwork/heimdall/blob/victor/pos-72-refactor-ci-workflows-to-publish-new/bridge/cmd/start.go#L53-L54, is it useful to start either of them without the other? can we remove service selection for bridge? Maybe we can do that in another PR

0xSasaPrsic:
hey @vcastellm  sure we can make adjustments but since this is already merged we can do it in another PR. Here is the part of the code where selection of processors is made based on the flag. https://github.com/maticnetwork/heimdall/blob/229673c44a43d925b546f7033a66c41fe638dec9/bridge/setu/processor/service.go#L83-L116

vcastellm:
@0xSasaPrsic I see, thanks for pointing me to that, I thought that the bridge had only 2 services, let's go with this as is.

PR #752, opened 2021-11-30
POS-53 Remove setup script for Heimdall and embed genesis files

igorcrevar:
We advise to users to use our Ansible scripts to deploy the Bor and Heimdall nodes using this playbook:

https://github.com/maticnetwork/node-ansible

There is a step in this process that runs a setup script for Heimdall to:

start the destination folder with $ heimdalld init

copy the cosmos genesis file to the folder, it is only available in the launch repo.

https://github.com/maticnetwork/node-ansible/blob/b7ab3f8078efd4a7337085fc68f672096f550708/roles/setup-network/tasks/setup.yml#L17

Unless this process is done, Heimdall cannot start.

The goal of this issue is to remove the need for this setup step.

Proposal

Move the code from the init command to its own function.

https://github.com/maticnetwork/heimdall/blob/master/cmd/heimdalld/init.go#L35

On starting the daemon, check if you need to run this script.

Embed in the Heimdall binary the cosmos-sdk genesis file for mainnet and mumbai chains.

This is for mainnet https://github.com/maticnetwork/launch/blob/master/mainnet-v1/without-sentry/heimdall/config/genesis.json

During the new init in the daemon, if the user does not supply any custom genesis file already, copy the prebuild embed genesis file to the destination folder (as the setup function does).

Notes

Do not remove the current init function, just move its functionality to the start command as well.

Since now we have the genesis file embed in the binary we might be tempted to load it from there instead of doing in it from a json file in the config directory, do not. Do not modify how the genesis file is being loaded.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@ $ make install
 ### Init-heimdall 
 ```bash 
 $ heimdalld init
+$ heimdalld init --chain=mainnet        Will init with genesis.json for mainnet
+$ heimdalld init --chain=mumbai         Will init with genesis.json for mumbai
 ```
 ### Run-heimdall 
 ```bash 
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/maticnetwork/heimdall
 
-go 1.12
+go 1.16
 
 require (
 	github.com/RichardKnop/machinery v1.7.4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v1
         with:
-          go-version: 1.13
+          go-version: 1.16
       - name: "Build binaries"
         run: make build
       - name: "Run tests"
--- a/cmd/heimdalld/init.go
+++ b/cmd/heimdalld/init.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -11,6 +12,7 @@ import (
 	"github.com/cosmos/cosmos-sdk/server"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	cfg "github.com/tendermint/tendermint/config"
 	"github.com/tendermint/tendermint/libs/cli"
 	"github.com/tendermint/tendermint/libs/common"
 	tmtime "github.com/tendermint/tendermint/types/time"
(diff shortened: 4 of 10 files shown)


igorcrevar:
Because decoding of embedded json files is done with
```
decoder := json.NewDecoder(f)
err = decoder.Decode(&result)
```
and `GenesisDoc` struct is defined in tendermint package and all `ConsensusParams` int fields do not accept string values,
mainnet.json and mumbai.json are copied from link provided in description, but those int fields are changed from string to int inside json (" is just removed). If those structs are defined in our package it would be easy to to change tag where needed or implement custom unmarshall. Other solution would be to create exactly same structs in our package but i do not think this is good (to have duplicate structs).

I hope current solution is ok(to remove quotes from int fields in json files)

README.md
```diff
@@ -16,6 +16,8 @@ $ make install
 ### Init-heimdall 
 ```bash 
 $ heimdalld init
+$ heimdalld init --chain=mainnet        Will init with genesis.json for mainnet
```
ferranbt, line 19: Is this still need it? I though we removed this and was integrated inside the daemon command.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/maticnetwork/heimdall
 
-go 1.12
+go 1.16
```
ferranbt, line 3: For any reviewer, we have bumped the Go version to use the new Go embed feature.

ferranbt (changes requested):

ferranbt:
> but those int fields are changed from string to int inside json

Isn't the genesis.json file generated from one of the cosmos-sdk command? So cosmos-sdk generates a command that then it cannot read? From what I understand this happens because we decode the json with: 
```
decoder := json.NewDecoder(f)
```
Is cosmos-sdk decoding the json file in another way and that is why it does not see this happening. Or is this a side effect of having the embed file?

README.md
igorcrevar: Yes, init will be executed through start command if genesis.json file does not exist. But i leave init command as it was before (for example if someone wants to init everything again for different network)

igorcrevar:
> I am not sure. I tried to use `_, err = cdc.UnmarshalBinaryLengthPrefixedReader(f, &result, 0)` inside `readPrealloc` but that failed with
"ERROR: unmarshal to types.GenesisDoc failed after 2 bytes (expected field number 1 <Varint> or field number 2 <Varint> , got 4): 0A20202267656E657369735F74696D65223A2022323032302D30352D33305430343A32383A30332E3137373035345A222C0A202022636861696E5F6964223A20226865696D64616C6C2D313337222C0A202022636F6E73656E7375735F706172616D73223A207B0A2020202022626C6F636B223A207B0A20202020"

so I figure out it would be best to use same code as code used in bor utils.go

igorcrevar:
> > but those int fields are changed from string to int inside json
> 
> Isn't the genesis.json file generated from one of the cosmos-sdk command? So cosmos-sdk generates a command that then it cannot read? From what I understand this happens because we decode the json with:
> 
> ```
> decoder := json.NewDecoder(f)
> ```
> 
> Is cosmos-sdk decoding the json file in another way and that is why it does not see this happening. Or is this a side effect of having the embed file?
I am not sure. I tried to use `_, err = cdc.UnmarshalBinaryLengthPrefixedReader(f, &result, 0)` inside `readPrealloc` but that failed with
"ERROR: unmarshal to types.GenesisDoc failed after 2 bytes (expected field number 1 <Varint> or field number 2 <Varint> , got 4): 0A20202267656E657369735F74696D65223A2022323032302D30352D33305430343A32383A30332E3137373035345A222C0A202022636861696E5F6964223A20226865696D64616C6C2D313337222C0A202022636F6E73656E7375735F706172616D73223A207B0A2020202022626C6F636B223A207B0A20202020"

so I figure out it would be best to use same code as code used in bor utils.go

go.mod
igorcrevar: Build is failing because build workflow uses 1.13 version eq:
```with:
          go-version: 1.13
```

ferranbt:
> Should I commit (or even ammend) this small fix?

Just commit the new change. I think this might be related to JSON encoding and decoding back and forth structs.

ferranbt:
Also, please fix the CI check.

igorcrevar:
> Also, please fix the CI check.

`ioutil` functions are deprecated (but still works) in go 1.16. https://www.srcbeat.com/2021/01/golang-ioutil-deprecated/
Do you want me to push `"io/ioutil"` replacements in 12 files ?
There are not too many changes eq:
- `ioutil.ReadAll(r.Body)` to `io.ReadAll(r.Body)` ,
- `outil.ReadDir(dbLocation) ` to `os.ReadDir(dbLocation)` ,
- `ioutil.WriteFile("priv_validator_key.json", jsonBytes, 0600)` to `os.WriteFile("priv_validator_key.json", jsonBytes, 0600)`
- `ioutil.TempFile("", "proposal")` to `os.CreateTemp("", "proposal")`
- `ioutil.TempDir("", dirPrefix)` to `os.MkdirTemp("", dirPrefix)`

ferranbt:
> Do you want me to push "io/ioutil" replacements in 12 files ?

Let's keep ioutil.

ferranbt (approved):

vcastellm (approved):
LGTM

PR #753, opened 2021-12-02
POS-86: Graceful shutdown of Heimdall deamon, REST and Bridge

momosh-ethernal:
- Graceful shutdown of Heimdall, REST, and Bridge are all initiated by the same context signaled by the OS on interrupt and program termination
- Necessary code duplications are due to the private functions and types from Cosmos SDK that needed changes to accommodate shutdowns

### DIFF
--- a/bridge/cmd/start.go
+++ b/bridge/cmd/start.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"os/signal"
@@ -19,6 +20,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/tendermint/tendermint/libs/common"
 	httpClient "github.com/tendermint/tendermint/rpc/client"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/maticnetwork/heimdall/helper"
 	"github.com/spf13/viper"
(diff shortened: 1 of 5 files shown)


vcastellm (commented):
Git history seems to have garbage, can you rebase with master?

momosh-ethernal:
@vcastellm not sure about master though, since this was done through V2 fork, hence all the noise.
@sh3ll3x3c agree, definetely

cmd/heimdalld/main.go
```diff
@@ -98,12 +102,14 @@ func GetSignerInfo(pub crypto.PubKey, priv []byte, cdc *codec.Codec) ValidatorAc
 
 func main() {
 	cdc := app.MakeCodec()
-	ctx := server.NewDefaultContext()
+	serverCtx := server.NewDefaultContext()
```
ferranbt, line 105: This variable change is creating a lot of new code.

server/root.go
```diff
@@ -24,35 +32,180 @@ import (
...
 
-	return err
+	return nil
 }
 
 // ServeCommands will generate a long-running rest server
 // (aka Light Client Daemon) that exposes functionality similar
 // to the cli, but over rest
-func ServeCommands(cdc *codec.Codec, registerRoutesFn func(*lcd.RestServer)) *cobra.Command {
+func ServeCommands(shutdownCtx ctx.Context, cdc *codec.Codec, registerRoutesFn func(ctx client.CLIContext, mux *mux.Router)) *cobra.Command {
```
ferranbt, line 201: I do not think we should pass this ```shutdownCtx``` to each command because the shutdown signal should be something specific to each execution of a command, this is, something to be created inside the ```Run``` command functions. 

Could we do:
``` 
RunE: func(cmd *cobra.Command, args []string) error {
	shutdownCtx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
	defer stop()
 	err := StartRestServer(shutdownCtx, cdc, registerRoutesFn, restCh)
}
```

server/root.go
```diff
@@ -24,35 +32,180 @@ import (
 	_ "github.com/maticnetwork/heimdall/server/statik"
 )
 
-func StartRestServer(cdc *codec.Codec, registerRoutesFn func(*lcd.RestServer), restCh chan struct{}) error {
-	rs := lcd.NewRestServer(cdc)
-	registerRoutesFn(rs)
+const shutdownTimeout = 10 * time.Second
+
+func StartRestServer(mainCtx ctx.Context, cdc *codec.Codec, registerRoutesFn func(ctx client.CLIContext, mux *mux.Router), restCh chan struct{}) error {
```
ferranbt, line 37: Is there any reason to call it ```mainCtx``` and not just ```ctx```?

server/root.go
```diff
@@ -66,11 +219,11 @@ func ServeCommands(cdc *codec.Codec, registerRoutesFn func(*lcd.RestServer)) *co
 }
 
 // RegisterRoutes register routes of all modules
-func RegisterRoutes(rs *lcd.RestServer) {
```
ferranbt, line 69: Why are we changing this?

ferranbt (changes requested):
Hey, it turns out that this "small" change include a lot of code changes with respect other PRs. I would like to propose another way to handle the graceful shutdown. 

Now, what you do is to have some "start" functions that setup the service (i.e. bridge, rest, daemon) and wait for it to finish. Then, you need to pass around the ```ctx``` so that they know when to stop.

Instead, something I usually do is for these "start" functions to only start (no wait to finish) and return some structs that export a ```Close``` function. Then, it is up to the functions creating and running the services (in this case, the cli commands) to start everything, wait for interrupt signal and Close every service. This ```Close``` function can be blocking and wait for it to finish everything.

server/root.go
momosh-ethernal: ok, done, fairly simple to change

server/root.go
momosh-ethernal: That RestServer struct is imported from Cosmos SDK. For us, to be able to shut down the REST server properly, we needed access to the listener field, which is private in this case.
Not copying the RestServer struct from the package gave us control, and as an added bonus we got that logger that was also private.

server/root.go
momosh-ethernal: In this case, each command would have created a context in its scope, where needed of course?

I could be wrong, but when we run our binary, each cmd should execute with the same context, since our main function is responsible for each execution of our defined commands?

Threading context down is a standard mechanism, not sure what I'm getting wrong here.

bridge/cmd/start.go
```diff
@@ -28,6 +30,105 @@ const (
 	waitDuration = 1 * time.Minute
 )
 
+// StartBridgeWithCtx starts bridge service and is able to shutdow gracefully
+// returns service errors, if any
+func StartBridgeWithCtx(shutdownCtx context.Context) error {
```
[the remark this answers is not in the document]
momosh-ethernal: The function signature is changed, StartBridgeWithCtx takes context instead of bool flag, and returns an error so it could be used with error groups properly.

StartBridge was left there since it's being used with the package init function, wrapped inside of a GetStartCmd. I think this is a support for older commands.

cmd/heimdalld/main.go
momosh-ethernal: This is passed to only two functions from the main itself, not sure I get this comment.

Also, if we could merge server context and standard context with context.WithValue, where server context would be nested.

0xSasaPrsic:
Hey @ferranbt  we spotted a bug during heimdall initialization so will make that fix in this PR.

0xSasaPrsic:
@ferranbt we made fix to heimdall initialization so it should be ready for review again.

ferranbt (approved):

PR #756, opened 2021-12-23
POS 32 Heimdall add seeds into code for mainnet

igorcrevar:
As part of the deployment of Heimdall in the PoS mainnet. The users have to use the seed/bootnode files defined in the Polygon docs to start the Heimdall process and connect to the network. This is not very UX friendly.

https://docs.polygon.technology/docs/validate/validate/run-validator-ansible/#configure-the-heimdall-service

Since the seed nodes do not change, this issue proposes to hardcode the values of this nodes inside the binary so that the user can run $ heimdall --chain mainnet and connect immediately to the network.

Otherwise, it would require to create a config file, check the docs to get the specific nodes and run Heimdall with that config. Note that this also make it difficult to run in automatic deployments.

Notes:

- If the user supplies other seed nodes either from flags (POS-17) or in the config file, we should use those seed nodes instead of the predefined ones

- unused function is deleted from server/root.go

### DIFF
--- a/helper/config.go
+++ b/helper/config.go
@@ -22,6 +22,7 @@ import (
 	logger "github.com/tendermint/tendermint/libs/log"
 	"github.com/tendermint/tendermint/privval"
 
+	cfg "github.com/tendermint/tendermint/config"
 	tmTypes "github.com/tendermint/tendermint/types"
 )
 
@@ -33,6 +34,7 @@ const (
 	RestServerFlag         = "rest-server"
 	BridgeFlag             = "bridge"
 	LogLevel               = "log_level"
+	SeedsFlag              = "seeds"
 
 	// heimdall-config flags
 	MainRPCUrlFlag               = "eth_rpc_url"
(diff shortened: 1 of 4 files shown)


helper/config.go
```diff
@@ -93,6 +95,8 @@ const (
 
 	DefaultTendermintNode = "tcp://localhost:26657"
 
+	DefaultMainnetSeeds string = "f4f605d60b8ffaaf15240564e58a81103510631c@159.203.9.164:26656,4fb1bc820088764a564d4f66bba1963d47d82329@44.232.55.71:26656,2eadba4be3ce47ac8db0a3538cb923b57b41c927@35.199.4.13:26656,3b23b20017a6f348d329c102ddc0088f0a10a444@35.221.13.28:26656,25f5f65a09c56e9f1d2d90618aa70cd358aa68da@35.230.116.151:26656"
```
vcastellm, line 98: Could we include `DefaultTestnetSeeds`?

vcastellm (changes requested):

helper/config.go
igorcrevar: We can, but in description of the task (in notes) there is this sentence: "Only use predefined seed nodes for mainnet since our mumbai testnet is only available to private partners.". If necessary, we can put a default value for Mumbai Testnet as well, but somebody has to provide me the default value.

vcastellm: AFAIK there's default seed nodes defined in the docs https://docs.polygon.technology/docs/integrate/full-node-deployment/#setup-full-node-for-testnetv4mumbai-testnet

Those should be correct

igorcrevar: default seeds for testnet is pushed (also unit test)

ferranbt (approved):

vcastellm (approved):

PR #757, opened 2021-12-28
Create codeql-analysis.yml

ssandeep:


vcastellm:
Errors are coming from swagger-ui I think we can dismiss them and merge?

vcastellm (approved):
LGTM

PR #760, opened 2022-01-07
Two stage build for Docker.

rootwarp:
This PR includes modification of `Dockerfile` to apply two stage build.
For running environment, Alpine linux is used and that improves security and reduces image size.

Unfortunately, Alpine linux does not include `glibc` package which is required for running Heimdall.
So `glibc` package is built by manually on second stage.

As a result, docker image size is reduced to 144MB compared to 2GB origin.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:latest
+FROM golang:latest AS builder
 
 ARG HEIMDALL_DIR=/heimdall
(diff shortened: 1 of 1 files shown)


vcastellm:
This and much more is already covered in https://github.com/maticnetwork/heimdall/pull/748, closing this, as new Docker images are coming soon. Thanks for your work!

PR #764, opened 2022-01-25
data limit size for clerk

temaniarpit27:


vcastellm (approved):

clerk/types/msg.go
```diff
@@ -59,6 +60,11 @@ func (msg MsgEventRecord) ValidateBasic() sdk.Error {
 	if msg.TxHash.Empty() {
 		return sdk.ErrInvalidAddress("missing tx hash")
 	}
+
+	if len(msg.Data) > helper.MaxStateSyncSize {
+		return ErrSizeExceed("length is larger than 100000")
```
anilCSE, line 65:
```suggestion
		return ErrSizeExceed("length is larger than 100000 bytes")
```
```suggestion
		return ErrSizeExceed("length is larger than 100000")
```

helper/config.go
```diff
@@ -69,6 +69,8 @@ const (
 	DefaultBorChainID string = "15001"
 
 	secretFilePerm = 0600
+
+	MaxStateSyncSize = 100000
```
anilCSE, line 73: Add a comment to this so it would be readable later

clerk/side_handler.go
```diff
@@ -90,12 +90,16 @@ func SideHandleMsgEventRecord(ctx sdk.Context, k Keeper, msg types.MsgEventRecor
...
 
 	if !bytes.Equal(eventLog.Data, msg.Data) {
-		k.Logger(ctx).Error(
-			"Data from event does not match with Msg Data",
-			"EventData", hmTypes.BytesToHexBytes(eventLog.Data),
-			"MsgData", hmTypes.BytesToHexBytes(msg.Data),
-		)
-		return hmCommon.ErrorSideTx(k.Codespace(), common.CodeInvalidMsg)
+		if len(eventLog.Data) > helper.MaxStateSyncSize && bytes.Equal(msg.Data, hmTypes.HexToHexBytes("")) {
+		} else {
```
anilCSE, line 94: Just use reverse condition and avoid else

anilCSE (commented):
This looks reasonable, but would love to see some tests.

bridge/setu/processor/clerk.go
```diff
@@ -95,6 +95,11 @@ func (cp *ClerkProcessor) sendStateSyncedToHeimdall(eventName string, logBytes s
 			"blockNumber", vLog.BlockNumber,
 		)
 
+		if len(event.Data) > helper.MaxStateSyncSize {
+			cp.Logger.Info("Data is too large", "data", hex.EncodeToString(event.Data))
+			event.Data = hmTypes.HexToHexBytes("")
```
ferranbt, line 100: Why are we resetting the data? Shouldn't we just return an error?

temaniarpit27: we can't return the error
if we return the error then there will be a nonce issue between the network and contracts which halt the state sync process
so we are resetting the data to "" so that we continue the state syncing

clerk/client/cli/tx.go
```diff
@@ -100,6 +101,11 @@ func CreateNewStateRecord(cdc *codec.Codec) *cobra.Command {
 				return fmt.Errorf("data should be hex string")
 			}
 
+			if len(data) > helper.MaxStateSyncSize {
+				fmt.Println("Data is too large", recordIDStr)
```
ssandeep, line 105: Let's change to warning log.

clerk/types/msg.go
```diff
@@ -59,6 +60,11 @@ func (msg MsgEventRecord) ValidateBasic() sdk.Error {
 	if msg.TxHash.Empty() {
 		return sdk.ErrInvalidAddress("missing tx hash")
 	}
+
+	if len(msg.Data) > helper.MaxStateSyncSize {
+		return ErrSizeExceed("length is larger than 100000 bytes")
```
ssandeep, line 65: Let's use the constant `MaxStateSyncSize` here.

clerk/client/cli/tx.go
ssandeep: And also add `resetting data to ""`

0xKrishna: Done

clerk/types/msg.go
0xKrishna: Done

0xKrishna (approved):

ssandeep (approved):

PR #779, opened 2022-03-01
Ensure `version` command runs on all binaries

ngotchac:
This PR ensures that the `version` command can be used with all binaries (`heimdalld`, `heimdallcli` and `bridge`).

One issue was that a config file was required in order to run any command (which is now only required when not running the `version` command).

Another issue was that the `BUILD_FLAGS` weren't passed to the `go build` command.


vcastellm (approved):
LGTM

jdkanani (approved):

PR #802, opened 2022-03-24
Added selected producers in state sync

0xsharma:
### [Linear Task](https://linear.app/matic/issue/POS-278/)

Use selected producers instead of all validators to send state-sync tx.

### DIFF
--- a/bridge/setu/processor/clerk.go
+++ b/bridge/setu/processor/clerk.go
@@ -48,6 +48,32 @@ func (cp *ClerkProcessor) RegisterTasks() {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


jdkanani:
FYI - This change won't send state sync if producers' bridges are down.

temaniarpit27:
> FYI - This change won't send state sync if producers' bridges are down.

Yes, but since we were thinking of solutions so that all validators don't send state sync, we thought maybe only the selected producers will send state sync. It will be around 20 so

temaniarpit27:
We want all the producers to emit state sync event

PR #804, opened 2022-03-28
fix: compare signer if no err occurs

ClarkChenc:
fix check operation in staking


ssandeep:
The condition in the current codebase is correct, no changes needed here.

PR #805, opened 2022-03-28
bridge/setu: check state sync tx against mempool while broadcast

manav2401:
This PR tries to achieve multiple things. 
1. It generalises the `isOldTx` function for all processors by moving it to base processor. 
2. Before broadcasting any state-sync transaction, it will also check with all existing transactions in the mempool through the `checkTxAgainstMempool` function in the base processor for duplicate transaction. This function is kept generic to allow enabling this for other transaction types as well. It will retry the same state-sync transaction post some fixed delay, if a transaction in mempool is found.

Checklist before merging
- [x] Code Implementation
- [x] Unit Tests (Checks for mock state-sync event and a single producer/node)
- [x] Testing in a multi-node devnet for real state-sync transactions
    - [x] Does it work for an actual state-sync transaction
    - [x] Does it work well with same state-sync transactions broadcasted by other peers and are lying in mempool
    - [x] Test the retry mechanism for transactions which encounter such condition
- [ ] Testing the same conditions on an internal node on mumbai and mainnet before release 
- [x] Revert changes in `broadcast_test.go`

Linear task
POS-277: [Mempool check for heimdall transactions](https://linear.app/matic/issue/POS-277/mempool-check-for-heimdall-transactions)

### DIFF
--- a/bridge/setu/processor/base.go
+++ b/bridge/setu/processor/base.go
@@ -106,7 +106,7 @@ func (bp *BaseProcessor) Stop() {
 
 // isOldTx checks if the transaction already exists in the chain or not
 // It is a generic function, which is consumed in all processors
-func (bp *BaseProcessor) isOldTx(cliCtx cliContext.CLIContext, txHash string, logIndex uint64, eventType string) (bool, error) {
+func (bp *BaseProcessor) isOldTx(cliCtx cliContext.CLIContext, txHash string, logIndex uint64, eventType util.BridgeEvent) (bool, error) {
 	queryParam := map[string]interface{}{
 		"txhash":   txHash,
 		"logindex": logIndex,
@@ -115,13 +115,13 @@ func (bp *BaseProcessor) isOldTx(cliCtx cliContext.CLIContext, txHash string, lo
 	// define the endpoint based on the type of event
 	var endpoint string
 	switch eventType {
-	case "staking":
+	case util.StakingEvent:
 		endpoint = helper.GetHeimdallServerEndpoint(util.StakingTxStatusURL)
-	case "topup":
+	case util.TopupEvent:
 		endpoint = helper.GetHeimdallServerEndpoint(util.TopupTxStatusURL)
-	case "clerk":
+	case util.ClerkEvent:
 		endpoint = helper.GetHeimdallServerEndpoint(util.ClerkTxStatusURL)
-	case "slashing":
+	case util.SlashingEvent:
 		endpoint = helper.GetHeimdallServerEndpoint(util.SlashingTxStatusURL)
 	}
 	url, err := util.CreateURLWithQuery(endpoint, queryParam)
--- a/bridge/setu/processor/base.go
+++ b/bridge/setu/processor/base.go
@@ -137,7 +137,7 @@ func (bp *BaseProcessor) isOldTx(cliCtx cliContext.CLIContext, txHash string, lo
 // checkTxAgainstMempool checks if the transaction is already in the mempool or not
 // It is consumed only for `clerk` processor
 func (bp *BaseProcessor) checkTxAgainstMempool(msg types.Msg) (bool, error) {
-	endpoint := helper.GetConfig().TendermintRPCUrl + "/unconfirmed_txs"
+	endpoint := helper.GetConfig().TendermintRPCUrl + util.TendermintUnconfirmedTxsURL
 	resp, err := http.Get(endpoint)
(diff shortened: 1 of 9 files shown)


0xKrishna (approved):

bridge/setu/processor/base.go
```diff
@@ -96,3 +103,123 @@ func (bp *BaseProcessor) String() string {
...
+	// checking the type of transaction, the transaction hash and log index
+	// present in the data of transaction
+
+	status := false
+Loop:
+	for _, txn := range Response.Result.Txs {
+		// Tendermint encodes the transactions with base64 encoding. Decode it first.
+		txBytes, err := base64.StdEncoding.DecodeString(txn)
+		if err != nil {
+			continue
```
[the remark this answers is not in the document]
manav2401: My bad. Will update it. Also, was wondering if we should use `Error` log, as it might appear a lot. I think it can be replaced by a warning or debug log. CC @0xKrishna.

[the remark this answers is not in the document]
0xKrishna: I don't think it will appear a lot, Use error for this./
In normal we should be able to decode a mempool txn.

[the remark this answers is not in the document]
manav2401: Yes, by appearing a lot, I mean if something breaks, it will appear a lot all of a sudden. Also, this is an additional check which we're doing before broadcasting a transaction, so even if this fails (due to some un-certain error), it actually doesn't cause any harm to the process.

bridge/setu/processor/base.go
```diff
@@ -96,3 +103,123 @@ func (bp *BaseProcessor) String() string {
...
+	}
+
+	// Iterate over txs present in the mempool
+	// We can verify if the message we're about to send is present by
+	// checking the type of transaction, the transaction hash and log index
+	// present in the data of transaction
+
+	status := false
+Loop:
+	for _, txn := range Response.Result.Txs {
```
cffls, line 173: Do we have some data about how big the mempool usually are and how many incoming state syncs events can burst up to? Getting all Txs from mempool and check every sync event against them with retry could have non-trivial impact on the performance (I could be wrong if the numbers are small).

0xKrishna: Usually mempool size is < 10 txns and we get up to 10-12 state sync events in a block, Max we can process as of now is 25 per Heimdall block. Your point is correct if we already have a huge number of txns in mempool but this check is actually implemented to prevent that only. We will also take a look at how we can improve this further.

manav2401: One thought which popped up my mind while implementing was to keep a map in memory as a cache for mempool transactions in a very specific format, which we require for the check. We can find any duplicate transaction through it in O(1).

cffls: Good idea about the memory cache. I am fine with current implementation. We can revisit this and optimize it if performance got hit in the future.

manav2401: Sure. Resolving this for now.

bridge/setu/processor/base.go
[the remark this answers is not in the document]
manav2401: Have made required changes. Resolving this.

temaniarpit27:
@manav2401 I hope that you have checked if the tx count is large, the API returns you all the transactions or is pagination required

bridge/setu/processor/staking.go
```diff
@@ -73,7 +73,7 @@ func (sp *StakingProcessor) sendValidatorJoinToHeimdall(eventName string, logByt
 		if len(signerPubKey) == 64 {
 			signerPubKey = util.AppendPrefix(signerPubKey)
 		}
-		if isOld, _ := sp.isOldTx(sp.cliCtx, vLog.TxHash.String(), uint64(vLog.Index)); isOld {
+		if isOld, _ := sp.isOldTx(sp.cliCtx, vLog.TxHash.String(), uint64(vLog.Index), "staking"); isOld {
```
0xKrishna, line 76: I can see `staking` hardcoded at multiple places in this file
Let's create a global variable in this file with `eventType` and use that so that it's easy to update if required, And use the same in switch case to compare. Will look better
Do this for all event types.

manav2401: Sure. This sounds better. Will do the required changes.

0xKrishna (approved):

0xsharma (approved):

PR #813, opened 2022-04-10
Use fees for each transaction rather than MaxGasFee (Dynamic Gas Fee)

temaniarpit27:


auth/types/stdtx.go
```diff
@@ -11,11 +12,14 @@ import (
 
 var (
 	_ sdk.Tx = (*StdTx)(nil)
+
+	maxGasWanted = uint64((1 << 63) - 1)
```
temaniarpit27, line 16: maxGasWanted should be max 1mil

auth/types/params.go
```diff
@@ -221,12 +173,6 @@ func (p Params) Validate() error {
 	if err := validateTxSizeCostPerByte(p.TxSizeCostPerByte); err != nil {
 		return err
 	}
-	if err := validateMaxTxGas(p.MaxTxGas); err != nil {
```
temaniarpit27, line 224: add these checks back

bor/keeper.go
```diff
@@ -92,8 +93,14 @@ func (k *Keeper) AddNewSpan(ctx sdk.Context, span hmTypes.Span) error {
 		return err
 	}
 
+	spanKey := GetSpanKey(span.ID)
```
temaniarpit27, line 96: revert these

helper/util.go
```diff
@@ -798,3 +828,61 @@ func FetchFromAPI(cliCtx cliContext.CLIContext, URL string) (result rest.Respons
...
+// CalculateGas simulates the execution of a transaction and returns
+// both the estimate obtained by the query and the adjusted amount.
+func CalculateGas(
+	queryFunc func(string, []byte) ([]byte, int64, error), cdc *codec.Codec,
+	txBytes []byte, adjustment float64,
+) (estimate, adjusted uint64, err error) {
+
+	// run a simulation (via /app/simulate query) to
+	// estimate gas and update TxBuilder accordingly
+	rawRes, _, err := queryFunc("/app/simulate", txBytes)
```
temaniarpit27, line 863: if gas is more than 1m, dont send with adjustment value, send max of 1m gas limit

PR #816, opened 2022-04-13
Change TaskDelayBetweenEachVal in bridge based on mempool

0xKrishna:
This PR adds additional delay by bridge in sending txs to heimdall based on the state of the mempool.

It will increase the delay factor for every 1000 unconfirmed txns in mempool.
For example if the current default delay is 12 Seconds
Then for upto 1000 txns it will stay as 12 sec
For 1000-2000 It will be 24 seconds
For 2000-3000 it will be 36 seconds

Basically for every 1000 txns it will increase the factor by 1.

### DIFF
--- a/bridge/setu/util/common.go
+++ b/bridge/setu/util/common.go
@@ -65,6 +65,8 @@ const (
 	RetryTaskDelay          = 12 * time.Second
 	RetryStateSyncTaskDelay = 24 * time.Second
 
+	mempoolTxnCountDivisor = 1000
+
 	// Bridge event types
 	StakingEvent  BridgeEvent = "staking"
 	TopupEvent    BridgeEvent = "topup"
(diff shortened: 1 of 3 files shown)


bridge/setu/util/common.go
```diff
@@ -173,8 +176,12 @@ func CalculateTaskDelay(cliCtx cliContext.CLIContext) (bool, time.Duration) {
 		}
 	}
 
+	// Change calculation later as per the discussion
```
ssandeep, line 179: Can you explain briefly the logic used for delay?

0xKrishna: Currently it will multiply delay for every 1000 unconfirmed txns in mempool
For example if the current default delay is 12 Seconds
Then for upto 1000 txns it will stay as 12 only
For 1000-2000 It will be 24 seconds
For 2000-3000 it will be 36 seconds

Basically for every 1000 txns it will increase the factor by 1.

ssandeep: Can you also add it as comment in the code?

bridge/setu/util/common.go
```diff
@@ -173,8 +176,12 @@ func CalculateTaskDelay(cliCtx cliContext.CLIContext) (bool, time.Duration) {
...
+	// Currently it will multiply delay for every 1000 unconfirmed txns in mempool
+	mempoolFactor := GetUnconfirmedTxnCount() / 1000
```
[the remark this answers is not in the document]
0xKrishna: On it

bridge/setu/util/common.go
0xKrishna: Okay

0xKrishna: Done

bridge/setu/util/common.go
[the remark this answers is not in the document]
0xKrishna: Done

0xsharma (approved):
LGTM !

manav2401 (approved):
LGTM

PR #820, opened 2022-04-27
[Bug fix] Add --all flag to bridge

cffls:
We need to provide flag `--all` to enable all bridge services. Otherwise the cli will fail
immediately.

### DIFF
--- a/builder/files/heimdalld.service
+++ b/builder/files/heimdalld.service
@@ -8,7 +8,7 @@
   WorkingDirectory=/usr/local/bin
   ExecStart=/usr/local/bin/heimdalld start --home /var/lib/heimdall \
     --chain=mainnet \
-    --bridge \
+    --bridge --all \
     --rest-server
   Type=simple
   User=root


builder/files/heimdalld.service
```diff
@@ -8,7 +8,7 @@
   WorkingDirectory=/usr/local/bin
   ExecStart=/usr/local/bin/heimdalld start --home /var/lib/heimdall \
```
cffls, line 9: @vcastellm Do we want to change home dir to `$HOME/.heimdalld`, which is currently used in [launch](https://github.com/maticnetwork/launch/blob/master/testnet-v4/sentry/validator/heimdall/setup.sh#L4)?

cffls, line 9: @vcastellm Do we want to change home dir to `$HOME/.heimdalld`, which is currently used in [launch](https://github.com/maticnetwork/launch/blob/master/testnet-v4/sentry/validator/heimdall/setup.sh#L4)?

cffls: I see, `$HOME` is not resolvable in system service file.

cffls: I see, `$HOME` is not resolvable in system service file.

PR #826, opened 2022-05-17
Docker changes for v0.3.0

cffls:
Change default heimdall home to `/var/lib/heimdall`. Remove initialization step from entry point since heimdalld is able to initialize configs automatically.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,13 @@
 FROM golang:latest
 
-ARG HEIMDALL_DIR=/heimdall
+ARG HEIMDALL_DIR=/var/lib/heimdall
 ENV HEIMDALL_DIR=$HEIMDALL_DIR
 
 RUN apt-get update -y && apt-get upgrade -y \
     && apt install build-essential git -y \
-    && mkdir -p /heimdall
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,13 +1,13 @@
...
+ARG HEIMDALL_DIR=/var/lib/heimdall
 ENV HEIMDALL_DIR=$HEIMDALL_DIR
 
 RUN apt-get update -y && apt-get upgrade -y \
     && apt install build-essential git -y \
-    && mkdir -p /heimdall
+    && mkdir -p $HEIMDALL_DIR
 
-WORKDIR ${HEIMDALL_DIR}
+WORKDIR /heimdall
```
temaniarpit27, line 10: shouldn't work dir be ${HEIMDALL_DIR}?

temaniarpit27 (approved):

Dockerfile
cffls: No, `HEIMDALL_DIR` will contain heimdall config and data, while `/heimdall` will only contain heimdall code.

vcastellm (commented):
Late LGTM this was here to maintain compatibility with old version

PR #830, opened 2022-06-02
Mardizzone/pos 452 add logging in datadog for state sync

marcello33:
Deployed and tested on a `TestnetV4-validator-node-1`

The purpose of this PR is to log as much info as possible through the whole functions chain followed by _StateSynced_ events in heimdall `bridge`.

As discussed, given the impact of _opentelemetry_ setup (more development time, context propagation, additional infrastructure work, DataDog costs), I started with simple logs to track functions and their execution time, to identify the bottleneck in heimdall bridge. 
We can deploy on our nodes and start collecting metrics on datadog, and in the meantime implement _opentelemetry_ - if needed - at least in the heimdall `bridge`. 

All the logs report the label `StateSyncedEvent`, the function name being called, the id of the _StateSynced_ event and the time elapsed within the function in milliseconds, e.g.

```
logger.Info("StateSyncedEvent: "+functionName,
			"stateSyncId", event.Id,
			"timeElapsed", time.Now().Sub(startTime).Milliseconds())
```

image


bridge/setu/broadcaster/broadcaster.go
```diff
@@ -57,9 +58,10 @@ func NewTxBroadcaster(cdc *codec.Codec) *TxBroadcaster {
 }
 
 // BroadcastToHeimdall broadcast to heimdall
-func (tb *TxBroadcaster) BroadcastToHeimdall(msg sdk.Msg) error {
+func (tb *TxBroadcaster) BroadcastToHeimdall(msg sdk.Msg, event interface{}) error {
 	tb.heimdallMutex.Lock()
 	defer tb.heimdallMutex.Unlock()
+	defer util.LogElapsedTimeForStateSyncedEvent(event, "BroadcastToHeimdall", time.Now())
```
JekaMas, line 64: Could we change LogElapsedTimeForStateSyncedEvent to receive only 1 type?

bridge/setu/util/common.go
```diff
@@ -476,3 +479,30 @@ func GetUnconfirmedTxnCount() int {
 
 	return count
 }
+
+// LogElapsedTimeForStateSyncedEvent logs useful info for StateSynced events
+func LogElapsedTimeForStateSyncedEvent(event interface{}, functionName string, startTime time.Time) {
```
JekaMas, line 484: Could we simplify it a bit? https://go.dev/play/p/OGwEm55IUs_z

bridge/setu/util/common.go
```diff
@@ -476,3 +479,30 @@ func GetUnconfirmedTxnCount() int {
...
+	switch event.(type) {
+	case statesender.StatesenderStateSynced:
+		event, ok := event.(statesender.StatesenderStateSynced)
+		if !ok {
+			logger.Error("Unable to typecast event to statesender.StatesenderStateSynced")
+			break
+		}
+		logger.Info("StateSyncedEvent: "+functionName,
+			"stateSyncId", event.Id,
```
JekaMas, line 493: We just need an ID here. Could we then pass everywhere not an event type, but rather eventID?

bridge/setu/broadcaster/broadcaster.go
marcello33: Not sure if that's worth, as - since the method is invoked also by functions using other event types - I would need to check if the type is `statesender.StatesenderStateSynced` in many other places

bridge/setu/util/common.go
marcello33: Same as other comment, if I am not mistaken I'd need to check if the type is `statesender.StatesenderStateSynced` in many other places and then typecast it to that event to extract the ID

JekaMas (approved):

bridge/setu/listener/rootchain.go
```diff
@@ -229,14 +232,19 @@ func (rl *RootChainListener) queryAndBroadcastEvents(rootchainContext *RootChain
...
-						rl.sendTaskWithDelay("sendUnstakeInitToHeimdall", selectedEvent.Name, logBytes, delay)
+						rl.sendTaskWithDelay("sendUnstakeInitToHeimdall", selectedEvent.Name, logBytes, 0, event)
+					} else if isCurrentValidator, delay := util.CalculateTaskDelay(rl.cliCtx, event); isCurrentValidator {
+						rl.sendTaskWithDelay("sendUnstakeInitToHeimdall", selectedEvent.Name, logBytes, delay, event)
 					}
 
 				case "StateSynced":
-					if isCurrentValidator, delay := util.CalculateTaskDelay(rl.cliCtx); isCurrentValidator {
-						rl.sendTaskWithDelay("sendStateSyncedToHeimdall", selectedEvent.Name, logBytes, delay)
+					event := new(statesender.StatesenderStateSynced)
```
temaniarpit27, line 241: why are we unpacking here?

bridge/setu/listener/rootchain.go
```diff
@@ -229,14 +232,19 @@ func (rl *RootChainListener) queryAndBroadcastEvents(rootchainContext *RootChain
...
+					if err := helper.UnpackLog(rl.stateSenderAbi, event, selectedEvent.Name, &vLog); err != nil {
```
temaniarpit27, line 242: why unpack here

bridge/setu/util/common.go
```diff
@@ -152,7 +153,8 @@ func IsInProposerList(cliCtx cliContext.CLIContext, count uint64) (bool, error)
 
 // CalculateTaskDelay calculates delay required for current validator to propose the tx
 // It solves for multiple validators sending same transaction.
-func CalculateTaskDelay(cliCtx cliContext.CLIContext) (bool, time.Duration) {
+func CalculateTaskDelay(cliCtx cliContext.CLIContext, event interface{}) (bool, time.Duration) {
```
temaniarpit27, line 156: get time for unconfirmed txns as well

temaniarpit27: @0xKrishna do you think adding mempool check will increase the delay time by a lot?

marcello33: Done, thank you

temaniarpit27 (approved):

bridge/setu/util/common.go
0xKrishna: It should only come into the picture when the txn is already in mempool, And if it's in mempool that means some bridge has already processed it and should get included in a block soon, But yeah we should add logs here as well to confirm if this is causing unnecessary delay.

PR #833, opened 2022-06-08
Mardizzone/pos 494 add flag to use json logger

marcello33:
This PR provides the possibility to use a JSON logger for `heimdall` application.
It requires the param `logs_type = "json"` in the `heimdall-config.toml` file
When the param value is different from _"json"_ or not provided, the app will default to the fallback logger used so far.
All the modules in `heimdall` will use such logger, provided from the `helper`
However, since the app underlying uses `cosmos-sdk`, some parts of the app (e.g. `consensus` and `rest-server`) will still log as usual.
For example - see comment in `server/root.go` - we could instantiate our own `RestServer`, but we would lose the dependency from the the sdk (not worth imho, but open to suggestions).

### DIFF
--- a/helper/config.go
+++ b/helper/config.go
@@ -202,7 +202,7 @@ func InitHeimdallConfigWith(homeDir string, heimdallConfigFilePath string) {
 	}
 
 	// perform check for json logging
-	if strings.Compare(conf.LogsType, "json") == 0 {
+	if conf.LogsType == "json" {
 		Logger = logger.NewTMJSONLogger(logger.NewSyncWriter(os.Stdout))
 	} else {
 		// default fallback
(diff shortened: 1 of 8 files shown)


temaniarpit27 (dismissed):

0xKrishna (dismissed):

helper/config.go
```diff
@@ -196,6 +201,14 @@ func InitHeimdallConfigWith(homeDir string, heimdallConfigFilePath string) {
 		log.Fatalln("Unable to unmarshall config", "Error", err)
 	}
 
+	// perform check for json logging
+	if strings.Compare(conf.LogsType, "json") == 0 {
```
JekaMas, line 205: Why not `conf.LogsType == "json"`?

JekaMas, line 205: Why not `conf.LogsType == "json"`?

temaniarpit27 (approved):

PR #841, opened 2022-06-21
Added metrics to self-healing logic

begmaroman:
Metrics endpoint: `...:2112/metrics`


JekaMas (dismissed):

0xKrishna (approved):

JekaMas (approved):

temaniarpit27:
@begmaroman have you checked this on one of the machines that the data if going to Datadog and you are able to view it and apply all kinds of filters?

begmaroman:
@temaniarpit27 I've tested it locally only. From my understanding, we will need to pull metrics from `/metrics` endpoint and then be able to create dashboards. Am I right here?

temaniarpit27:
I think these metrics should also be captured in datadog
we might need to change the datadog agent file but not sure
@vcastellm can you please comment?

begmaroman:
@temaniarpit27 I see we are doing the same on the bor side but using opentel. The HTTP endpoint that exposes metrics data is the same. Seems like we need to adjust datadog agent. Not sure if it could be done within this PR.

begmaroman:
@temaniarpit27 as discussed in Slack, we need to add the daemon metrics config to this list: https://github.com/maticnetwork/ansible/blob/master/playbooks/datadog_agent_blockchain_nodes.yml#L59. Currently we are getting metrics from the heimdall daemon and bor, but not bridge since it does not have any metrics apart from ones defined within this PR.

temaniarpit27:
> @temaniarpit27 as discussed in Slack, we need to add the daemon metrics config to this list: https://github.com/maticnetwork/ansible/blob/master/playbooks/datadog_agent_blockchain_nodes.yml#L59. Currently we are getting metrics from the heimdall daemon and bor, but not bridge since it does not have any metrics apart from ones defined within this PR.

Can we create a linear task for this and inform devops team so that they can pick this up?

temaniarpit27 (approved):

.github/workflows/ci.yml
```diff
@@ -5,16 +5,20 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
+
       - name: Install Go
-        uses: actions/setup-go@v1
+        uses: actions/setup-go@v3
         with:
-          go-version: 1.13
+          go-version: 1.17.x
```
vcastellm, line 13: can we use `1.18.x` as in bor?

vcastellm (approved):
LGTM aside from the Go version

begmaroman:
@temaniarpit27 done, shared a talk link in Slack

.github/workflows/ci.yml
begmaroman: Done

temaniarpit27 (approved):

vcastellm (approved):

PR #857, opened 2022-07-18
Implemented gRPC for heimdall-bor interaction

0xKrishna:
In this PR we have

1. Added a gRPC server within the rest-server which is listening at port 3132 by default
2. Using https://github.com/maticnetwork/polyproto/releases/tag/v0.0.1

This should ideally replicate https://github.com/maticnetwork/bor/blob/develop/consensus/bor/heimdall.go#L12

https://polygon.atlassian.net/browse/POS-647


server/proto/heimdall.proto
```diff
@@ -0,0 +1,78 @@
...
+    rpc GetEventRecords(GetEventRecordsRequest) returns (stream GetEventRecordsResponse) {}
+}
+
+message Validator {
+    uint64 ID = 1;
+    uint64 startEpoch = 2;
+    uint64 endEpoch = 3;
+    uint64 nonce = 4;
+    int64 power = 5;
+    string pubKey = 6;
```
JekaMas, line 21: can be use more strict type for it?

server/proto/heimdall.proto
```diff
@@ -0,0 +1,78 @@
...
+    string signer = 7;
+    string last_updated = 8;
```
JekaMas, line 23: don't we have timestamp type to be used here?

server/proto/heimdall.proto
```diff
@@ -0,0 +1,78 @@
...
+message Span {
+    uint64 span_id = 1;
+    uint64 start_block = 2;
+    uint64 end_block = 3;
+    ValidatorSet validator_set = 4;
+    repeated Validator selected_producers = 5;
+    string bor_chain_id = 6;
+}
+
+message GetLatestSpanRequest {
```
JekaMas, line 42: google.protobuf.Empty ?

server/proto/heimdall.proto
```diff
@@ -0,0 +1,78 @@
...
+option go_package = "/server/proto";
+
+service Heimdall {
+    rpc GetLatestSpan(GetLatestSpanRequest) returns (GetLatestSpanResponse) {}
+    rpc GetSpan(GetSpanRequest) returns (GetSpanResponse) {}
+    rpc GetEventRecords(GetEventRecordsRequest) returns (stream GetEventRecordsResponse) {}
+}
+
+message Validator {
+    uint64 ID = 1;
```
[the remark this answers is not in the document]
JekaMas: In Bor we have an interface

```
	GetCurrentSpan(ctx context.Context, headerHash common.Hash) (*span.Span, error)
	GetCurrentValidators(ctx context.Context, headerHash common.Hash, blockNumber uint64) ([]*valset.Validator, error)
	CommitSpan(ctx context.Context, heimdallSpan span.HeimdallSpan, state *state.StateDB, header *types.Header, chainContext core.ChainContext) error
```

does it match to that proto file?

server/gRPC.go
```diff
@@ -0,0 +1,152 @@
...
+const (
+	LatestSpanURL   = "/bor/latest-span"
+	SpanURL         = "/bor/span/%v"
+	EventRecordList = "/clerk/event-record/list"
+)
+
+var logger tmLog.Logger
+
+func setupGRPCServer(shutDownCtx context.Context, cdc *codec.Codec, addr string, lggr tmLog.Logger) error {
+	logger = lggr
```
JekaMas, line 27: Love it! It is a great first step to have gRPC.

Could we have https://github.com/grpc-ecosystem/go-grpc-middleware with proper middlewares as next?

server/proto/heimdall.proto
0xKrishna: Checking

server/gRPC.go
0xKrishna: Sure, Working on all the above comments,Thanks for the review.

server/proto/heimdall.proto
0xKrishna: Okay, I used string because the `last_updated` we have here is in UNIX timestamp format
https://heimdall.api.matic.today/staking/validator/1

In the other `EventRecord` message of this proto I have used `google.protobuf.Timestamp` because it has normal timestamp format
https://heimdall.api.matic.today/clerk/event-record/36452

JekaMas: Could we convert the string timestamp into time.Time or uint64 for Heimdall->Bor interaction and use google.protobuf.Timestamp ?

0xKrishna: Sure, Will check that and test.

server/proto/heimdall.proto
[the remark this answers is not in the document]
JekaMas: Could you specify what Bor interfaces are touched by this Heimdall changes?

server/proto/heimdall.proto
```diff
@@ -0,0 +1,78 @@
+syntax = "proto3";
+
+package proto;
+
+import "google/protobuf/timestamp.proto";
+
+option go_package = "/server/proto";
+
+service Heimdall {
+    rpc GetLatestSpan(GetLatestSpanRequest) returns (GetLatestSpanResponse) {}
```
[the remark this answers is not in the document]
0xKrishna: Yeah, All these methods are Bor requesting data from heimdall
1. Span -> It's a one time single object response, Hence no need of stream
2. Event -> This is where I have used stream as we query events in a loop, Now they can be simply streamed using this.

server/proto/heimdall.proto
0xKrishna: I checked and found that this is not even used in Bor
https://github.com/maticnetwork/bor/blob/88dbfa1c13c15464d3c1a3085a9f12d0ffb9b218/consensus/bor/heimdall/span/span.go#L8
Can we ignore the values which are not consumed by Bor then ?

server/proto/heimdall.proto
0xKrishna: Removed

server/proto/heimdall.proto
0xKrishna: Removed

server/proto/heimdall.proto
```diff
@@ -0,0 +1,78 @@
...
+    Validator proposer = 2;
+}
+
+message Span {
+    uint64 span_id = 1;
+    uint64 start_block = 2;
+    uint64 end_block = 3;
+    ValidatorSet validator_set = 4;
+    repeated Validator selected_producers = 5;
+    string bor_chain_id = 6;
```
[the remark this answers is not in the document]
0xKrishna: Even on Bor side this is referred as string, So keeping it as it is for now, Let me know if you want me to change.

server/proto/heimdall.proto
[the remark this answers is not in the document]
0xKrishna: Updated

server/proto/heimdall.proto
0xKrishna: Done

server/proto/heimdall.proto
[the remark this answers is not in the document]
0xKrishna: https://github.com/maticnetwork/bor/blob/master/consensus/bor/rest.go#L26

server/gRPC.go
0xKrishna: I have added one interceptor based on what I saw in bor code, Looking to add more like prometheus.

0xKrishna: Didn't found much use of any other as of now, Will review middleware part later once integrated from bor side as well.

JekaMas (approved):

JekaMas (approved):

manav2401 (approved):

PR #858, opened 2022-07-21
new: dev: pos-621 replace encoding/json with jsoniter

marcello33:
This PR replaces the json standard library `encoding/json` with the faster alternative `jsoniter`

The only missing replacement is due to the dependency to the `amino.Codec` of `cosmos-sdk` and `tendermint`, which still uses the methods `cdc.MustMarshalJSON(...)` and `cdc.MustUnmarshalJSON(...)`

Benchmarks are provided within the PR. Here the results:
```
BenchmarkJsonStandardLibrary-12    	                   10129   	     111009 ns/op	   35584 B/op	     376 allocs/op
BenchmarkJsoniterLibraryWithDefaultConfig-12    	   18061	     64152 ns/op	   40466 B/op	     552 allocs/op
BenchmarkJsoniterLibraryWithFastestConfig-12    	   18472	     62357 ns/op	   39174 B/op	     392 allocs/op
```

As claimed by the [official doc](https://github.com/json-iterator/go), the ns/op are significantly reduced (in this case 63533 vs 114246)

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -503,7 +503,7 @@ func (app *HeimdallApp) Name() string { return app.BaseApp.Name() }
 func (app *HeimdallApp) InitChainer(ctx sdk.Context, req abci.RequestInitChain) abci.ResponseInitChain {
 	var genesisState GenesisState
 
-	var json = jsoniter.ConfigCompatibleWithStandardLibrary
+	json := jsoniter.ConfigCompatibleWithStandardLibrary
 
 	if err := json.Unmarshal(req.AppStateBytes, &genesisState); err != nil {
 		panic(err)
--- a/app/app.go
+++ b/app/app.go
@@ -502,7 +502,8 @@ func (app *HeimdallApp) Name() string { return app.BaseApp.Name() }
 // InitChainer initializes chain
 func (app *HeimdallApp) InitChainer(ctx sdk.Context, req abci.RequestInitChain) abci.ResponseInitChain {
 	var genesisState GenesisState
-	var json = jsoniter.ConfigCompatibleWithStandardLibrary
+
+	json := jsoniter.ConfigCompatibleWithStandardLibrary
 
 	if err := json.Unmarshal(req.AppStateBytes, &genesisState); err != nil {
 		panic(err)
--- a/app/app.go
+++ b/app/app.go
@@ -503,7 +503,7 @@ func (app *HeimdallApp) Name() string { return app.BaseApp.Name() }
 func (app *HeimdallApp) InitChainer(ctx sdk.Context, req abci.RequestInitChain) abci.ResponseInitChain {
 	var genesisState GenesisState
 
-	if err := jsoniter.ConfigFastest.Unmarshal(req.AppStateBytes, &genesisState); err != nil {
+	if err := jsoniter.ConfigCompatibleWithStandardLibrary.Unmarshal(req.AppStateBytes, &genesisState); err != nil {
 		panic(err)
 	}
 
--- a/app/app.go
+++ b/app/app.go
@@ -1,13 +1,13 @@
 package app
 
 import (
-	"encoding/json"
 	"fmt"
 
 	bam "github.com/cosmos/cosmos-sdk/baseapp"
 	"github.com/cosmos/cosmos-sdk/codec"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/module"
+	jsoniter "github.com/json-iterator/go"
 	abci "github.com/tendermint/tendermint/abci/types"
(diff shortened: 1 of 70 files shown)


app/app.go
```diff
@@ -502,6 +502,9 @@ func (app *HeimdallApp) Name() string { return app.BaseApp.Name() }
 // InitChainer initializes chain
 func (app *HeimdallApp) InitChainer(ctx sdk.Context, req abci.RequestInitChain) abci.ResponseInitChain {
 	var genesisState GenesisState
+
+	json := jsoniter.ConfigCompatibleWithStandardLibrary
```
JekaMas, line 506: Do we need a standard compatibility? Is it for all APIs or just bor<->heimdall? If the second option I'd rather use the fastest config.

app/export.go
```diff
@@ -19,7 +20,8 @@ func (app *HeimdallApp) ExportAppStateAndValidators() (
 
 	// create app state
 	// appState, err = codec.MarshalJSONIndent(app.cdc, genState)
-	appState, err = json.Marshal(result)
+	var jsonLib = jsoniter.ConfigCompatibleWithStandardLibrary
```
JekaMas, line 23: Can we have one config somewhere and just reuse it?

auth/client/cli/query.go
```diff
@@ -1,13 +1,13 @@
...
 
 import (
-	"encoding/json"
 	"fmt"
 	"strings"
 
 	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/client/context"
 	"github.com/cosmos/cosmos-sdk/codec"
+	jsoniter "github.com/json-iterator/go"
```
JekaMas, line 10: Why not json alias? Why do we need this alias renaming in so many places?

app/app.go
marcello33: I left the most compatible choice overall.
Here the configs for fastest:
```
// ConfigFastest marshals float with only 6 digits precision
var ConfigFastest = Config{
	EscapeHTML:                    false,
	MarshalFloatWith6Digits:       true, // will lose precession
	ObjectFieldMustBeSimpleString: true, // do not unescape object field
}.Froze()
```

Wondering if this could break something in our APIs given the precision loss

JekaMas: We just need tests from Bor side and ready to go. Also we can use the same package from Bor which will give us 100% compatibility.

app/app.go
```diff
@@ -502,7 +502,8 @@ func (app *HeimdallApp) Name() string { return app.BaseApp.Name() }
 // InitChainer initializes chain
 func (app *HeimdallApp) InitChainer(ctx sdk.Context, req abci.RequestInitChain) abci.ResponseInitChain {
 	var genesisState GenesisState
-	if err := json.Unmarshal(req.AppStateBytes, &genesisState); err != nil {
+
+	if err := jsoniter.ConfigFastest.Unmarshal(req.AppStateBytes, &genesisState); err != nil {
```
JekaMas, line 506: Was the task only about bor and Heimdall API? Did you changed json package everywhere?

bridge/setu/processor/checkpoint.go
```diff
@@ -3,14 +3,14 @@ package processor
...
 	"encoding/hex"
-	"encoding/json"
 	"errors"
 	"math"
 	"math/big"
 	"strconv"
 	"time"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	jsoniter "github.com/json-iterator/go"
```
JekaMas, line 13: Do we store json anywhere in Heimdall into blockchain?

app/app.go
marcello33: Changed everywhere possible, except for cosmos and tendermint dependencies 
Integration tests are successful so I suppose no breaking changes have been introduced

bridge/setu/processor/checkpoint.go
marcello33: Not that I am aware of

app/app.go
JekaMas: Ok, let's give it a try. A bit risky, but still.

marcello33: Will inform also the team about the risks so that if we detect something strange, we know how to react

bridge/setu/processor/checkpoint.go
JekaMas: Ok. We'll see it in tests and deploy.

app/app.go
JekaMas: Thank you!

JekaMas (approved):

0xKrishna (approved):

JekaMas:
@marcello33 let's merge the PR and redo profiling then.

PR #862, opened 2022-07-28
Add 'heimdall' user during package installation

cffls:
This change will check and create a new 'heimdall' user when a heimdall binary is installed on a user's machine.

### DIFF
--- a/builder/files/heimdalld.service
+++ b/builder/files/heimdalld.service
@@ -8,10 +8,9 @@
   WorkingDirectory=/usr/local/bin
   ExecStart=/usr/local/bin/heimdalld start --home /var/lib/heimdall \
     --chain=mainnet \
-    --bridge --all \
     --rest-server
   Type=simple
-  User=ubuntu
   LimitNOFILE=65536
+  User=heimdall
 [Install]
   WantedBy=multi-user.target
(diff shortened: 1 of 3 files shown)


pratikpatil024 (approved):

builder/files/heimdalld.service
```diff
@@ -8,10 +8,9 @@
   WorkingDirectory=/usr/local/bin
   ExecStart=/usr/local/bin/heimdalld start --home /var/lib/heimdall \
     --chain=mainnet \
-    --bridge --all \
     --rest-server
   Type=simple
-  User=ubuntu
   LimitNOFILE=65536
+  User=heimdall
```
JekaMas, line 14: Do we need 2 users? Could we stick with just 1 user name? Like Bor?

JekaMas (approved):

builder/files/heimdalld.service
cffls: I think it is better to create two users, one for heimdall and one for bor. Using one user may be confusing. Also, one of the reasons to create new users is to get better file permission control.

JekaMas: Ok, in nearest future I'm planning to merge Heimdall and Bor, then will remove one of this users.

PR #876, opened 2022-09-18
bridge/setu/listener, helper: consume post-merge eth client for 'finalized' block

manav2401:
This PR updates the bor package to consume new ethclient, which helps in fetching `finalized` blocks on L1. They will act as the primary source of truth for block confirmations and will use `MainchainTxConfirmations` value as a fall back option. 

It also fixes a bug in the self-healing logic where empty event (with no topics) wasn't handled (Reasonings for the same is kept as a separate task). 

TODOs before merge
- [x] Merge/finalize bor release/branch and update go.mod if required
- [x] Tests
    - [x] Testing on Mumbai where goerli API's would provide us with "finalized" blocks
    - [x] Testing the fallback mechanism and general things (like checkpoint and span) on a devnet


bridge/setu/listener/base.go
```diff
@@ -169,10 +175,30 @@ func (bl *BaseListener) StartPolling(ctx context.Context, pollInterval time.Dura
...
+					// finalized was requested
+					bHeader = &blockHeader{header: header, isFinalized: true}
+				} else {
+					// latest was requested
+					bHeader = &blockHeader{header: header, isFinalized: false}
+				}
+			}
+
+			// if error occured and finalized was requested, fall back to latest block
+			if err != nil && number != nil {
```
JekaMas, line 192: Is the case `if err != nil && number == nil` possible? Should we cover it?

JekaMas: Shall we log the error in case it's not nil?

JekaMas (approved):
LGTM with few comments

bridge/setu/listener/base.go
manav2401: Well, in that case we don't need to send the data to the channel. 

I didn't cover it because it was the same earlier as well. See https://github.com/maticnetwork/heimdall/blob/develop/bridge/setu/listener/base.go#L172-L176.

manav2401: This is a case when the matic chain listener will call this function with `number` as `nil`, as we need the latest block and not the finalized block.

manav2401: Okay logging makes sense. Will do. Thanks!

JekaMas (approved):

Raneet10 (approved):

PR #884, opened 2022-09-23
Update tendermint version

cffls:
This change will bring in the [log tag change](https://github.com/maticnetwork/tendermint/pull/25) in tendermint.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,7 +143,7 @@ jobs:
           if: always()
           run: |
             cd matic-cli/devnet
-            docker compose down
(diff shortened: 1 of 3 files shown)


pratikpatil024 (approved):

temaniarpit27:
@cffls: Should we close this PR and after merging your PR to develop, we should raise the maticnetwork:develop branch against qa. I know the commits will be the same but I think it will be great to follow that path always to avoid any discrepancy. What do you think?

cffls:
@temaniarpit27 Thanks Arpit. This makes sense. I will close this one and cherry-pick it from qa after.

PR #895, opened 2022-10-27
OpenTelemetry Support

0xsharma:
In this PR, we are adding support for OpenTelemetry Collector. Developers will be able to trace multiple functions and send traces to observability dashboards like Datadog, Jaeger etc.

In this PR, we introduce two flags : 
- `--open-tracing` this boolean flag enables openTracing. [Default = false]
- `--open-collector-endpoint=<ENDPOINT>` this flag can be used to configure the otel-collector endpoint to send the traces. [Default="localhost:4317"]


JekaMas (approved):

bridge/setu/processor/clerk.go
```diff
@@ -88,6 +99,13 @@ func (cp *ClerkProcessor) sendStateSyncedToHeimdall(eventName string, logBytes s
 			)
 			return nil
 		}
+		tracing.EndSpan(isOldTxSpan)
```
manav2401, line 102: We'd need to end span in case of errors as well. If `isOldTx` returns an error, we would just exit the function. 

Maybe just do 
```
_, isOldTxSpan := tracing.StartSpan(sendStateSyncedToHeimdallCtx, "isOldTx")

isOld, err := cp.isOldTx(...);

tracing.EndSpan(isOldTxSpan)

if err != nil {...}
```

Same for `checkTxAgainstMempool`. 

Also, it would be good to add an `err` attribute printing the error if any.

bridge/setu/processor/clerk.go
```diff
@@ -88,6 +99,13 @@ func (cp *ClerkProcessor) sendStateSyncedToHeimdall(eventName string, logBytes s
...
+
+		tracing.SetAttributes(sendStateSyncedToHeimdallSpan, []attribute.KeyValue{
```
manav2401, line 104: Should we move this above `IsOldTx` calls?

PR #899, opened 2022-11-22
Created service package

0xKrishna:
# Description

Restructuting changes to heimdalld folder, So that we can start heimdall process from other services/scripts.
- `NewHeimdallService` as a constructor method to start the process
- `GetHeimdallApp` to get the heimdall app object in external services

We have also removed code dependency on bor repo, Moved all the required structs which we were using from bor repo to `bor/types` in heimdall repo itself.

# Changes

# Checklist

# Testing

- [ ] I have added unit tests
- [x] I have tested this code manually on local environment
- [x] I have tested this code manually on mumbai


bor/beginblocker.go
```diff
@@ -22,7 +21,7 @@ func BeginBlocker(ctx sdk.Context, _ abci.RequestBeginBlock, k Keeper) {
 			return
 		}
 
-		var spans []*bor.ResponseWithHeight
+		var spans []*types.ResponseWithHeight
```
JekaMas, line 24: why don't keep it in Bor package or, if needed, bor/api ?

JekaMas (approved):

bor/beginblocker.go
0xKrishna: That creates dependency issue, I tried.
Then moved it to heimdall repo and it worked.

bor/client/rest/query.go
```diff
@@ -534,13 +532,13 @@ func loadSpanOverrides() {
 		return
 	}
 
-	var spans []*bor.ResponseWithHeight
+	var spans []*types.ResponseWithHeight
```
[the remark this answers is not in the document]
0xKrishna: Okay

bor/beginblocker.go
JekaMas: Even a new package like bor/api?

0xKrishna: Okay, Let me try that

0xKrishna: While I was checking this I noticed that in latest bor code (0.3.x) we have removed `ResponseWithHeight` from the code, There is no struct like that anymore. I am looking more analyse the impact of that on heimdall code.

0xKrishna: So in latest version of bors we don't have these structs which I have replaced, So I think we will anyway have to update these after we release v0.3.0 on bor and update that in heimdall. I will create a task for that but for now I think the change I did should be fine.

PR #945, opened 2023-02-08
POS-1114 and POS-1180: fix UnstakeInit bug and use constants for smart contracts' events

marcello33:
# Description

This PR fixes the bug related to `UnstakeInit` and introduces constants to use when unpacking logs originated from smart contracts events instead of plain strings.

# Changes

# Checklist

## Testing

- [ ] I have added unit tests
- [ ] I have tested this code manually on mumbai
- [x] I have created new e2e tests into express-cli

### DIFF
--- a/bridge/setu/listener/rootchain_events.go
+++ b/bridge/setu/listener/rootchain_events.go
@@ -19,7 +19,9 @@ import (
 const (
 	blocksRange   = 1000
 	maxIterations = 1 // 1 iteration (1000 blocks) on rootchain is roughly 3.3 hours, which is enough for backtracking
+)
 
+const (
 	// smart contracts' events names
 	stateSyncedEvent = "StateSynced"
 	stakeUpdateEvent = "StakeUpdate"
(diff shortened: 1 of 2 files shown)


temaniarpit27:
@marcello33 we will be making the same mistake again. we need to make sure it goes via hardfork. otherwise if 50% upgrade and rest don't and someone does unstake, then the chain will halt

marcello33:
> @marcello33 we will be making the same mistake again. we need to make sure it goes via hardfork. otherwise if 50% upgrade and rest don't and someone does unstake, then the chain will halt

Right @temaniarpit27 
Will discuss this with the other on the process to include the changes in the next hard fork

marcello33:
@VAIBHAVJINDAL3012's taking care of the hardfork related stuff.

temaniarpit27:
@marcello33 can we atleast code the part where we are implementing the hard fork. that will avoid any confusions or mistakes later on. the block number can be hardcoded later but at least we can test it in dev env and keep the PR ready.

marcello33:
Yes @temaniarpit27 that's what I agreed with Vaibhav 🙌🏼

VAIBHAVJINDAL3012 (approved):

marcello33:
Included in https://github.com/maticnetwork/heimdall/pull/958

PR #993, opened 2023-05-25
clerk/client/rest :: add : blocktime check in recordListHandlerFn

0xsharma:
# Description

Handle `heimdalld not in sync` when bor is querying latest state-syncs for syncing blocks. In request we send 'to-time' timestamp. We return `nothing` if `to-time` > `heimdall_latest_block_timestamp`

### DIFF
--- a/clerk/client/rest/query.go
+++ b/clerk/client/rest/query.go
@@ -215,6 +215,21 @@ func recordListHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
 				return
 			}
 
+			node, err := cliCtx.GetNode()
+			if err != nil {
+				return
+			}
+
+			status, err := node.Status()
+			if err != nil {
+				return
+			}
+
+			time := status.SyncInfo.LatestBlockTime
+			if toTime > time.Unix() {
+				return
+			}
+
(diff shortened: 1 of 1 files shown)


clerk/client/rest/query.go
```diff
@@ -215,6 +215,21 @@ func recordListHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
...
+			}
+
+			status, err := node.Status()
+			if err != nil {
+				return
+			}
+
+			time := status.SyncInfo.LatestBlockTime
+			if toTime > time.Unix() {
+				return
```
cffls, line 230: Can we add a log or metric here for visibility?

0xsharma: A log wouldn't be necessary. Heimdall logs update crazy fast when its syncing. A user will be able to understand the same. What say ?

cffls: I was thinking how we can get alerted when this happens. Just realized that we have a warning message [in bor](https://github.com/maticnetwork/bor/blob/develop/consensus/bor/heimdall/client.go#L177) when it is not getting response from heimdall. I am fine with this PR then. Thanks!

PR #1003, opened 2023-06-10
Fix HaltHeight and HaltTime config/CLI options so that they work

svenski123:
The halt-height and halt-time options are useful to sync the heimdall daemon to a predetermined block height or block time and then exit, allowing state at a given point in time to be reproduced for research and debugging purposes.

The Cosmos SDK which is part of heimdall defines these configuration options and implements them in the BaseApp.

However heimdall does not pass the relevant configuration to NewBaseApp() resulting in these options being ignored; this pull request addresses this.

There is additionally one clean up replacing the use of the literal string "pruning" for retrieving a configuration option the constant flagPruning which is already defined in the source code file and is used elsewhere.

# Changes

- [X] Bugfix (non-breaking change that solves an issue)
- [X] New feature (non-breaking change that adds functionality)

### DIFF
--- a/cmd/heimdalld/service/service.go
+++ b/cmd/heimdalld/service/service.go
@@ -216,7 +216,10 @@ func getNewApp(serverCtx *server.Context) func(logger log.Logger, db dbm.DB, sto
 		helper.InitHeimdallConfig("")
 		helper.UpdateTendermintConfig(serverCtx.Config, viper.GetViper())
 		// create new heimdall app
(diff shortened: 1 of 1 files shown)


cffls (approved):
LGTM, thanks for the fix!

svenski123:
I opened this trivial 5 line diff about a month ago, all checks have passed, one positive comment and then ... nothing.

There are only six other open pull requests right now and half of them with "don't merge" in their title so there doesn't appear to be a lot of other activity.

Do you actually want bug fixes and contributions from the wider community?

0xKrishna:
@svenski123 Apologies for this situation, Taking a look, Will merge this on priority.

0xKrishna (approved):

PR #1017, opened 2023-07-07
feature: upgrade remote node client versions via ci

rebelArtists:
# Description

still POC: ensure we can use github actions CI flow to trigger AWX API call that ultimately updates a fleet of remote PoSV1 instances, this is on the back of RFC here: https://www.notion.so/polygontechnology/RFC-PoSV1-Test-Fleet-0f0821f9c97a48b9adf273e2e7a48a66?pvs=4

# Changes

- Add Github Actions workflow to call AWX API and update remote nodes heimdall version
- as part of the same workflow, periodically check (on loop) the health status of all relevant datadog monitors, indicating whether the proposed client changes have disrupted the network or all changes are stable (bake time)

### DIFF
--- a/.github/workflows/prerelease-test-fleet.yml
+++ b/.github/workflows/prerelease-test-fleet.yml
@@ -1,0 +1,30 @@
+name: prerelease-test-fleet
+
+on:
+  push:
+  # release:
+  #   types: [published]
+
+jobs:
+  prerelease-test-fleet:
+    # if: "github.event.release.prerelease"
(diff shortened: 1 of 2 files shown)


marcello33:
Hi @rebelArtists, thanks for your contribution.
Can you please fill all the sections of the PR template so that we could get a better understanding about what you're trying to achieve with such changes?
Thank you

rebelArtists:
poc for posv1 test fleet, may reopen in future

PR #1037, opened 2023-08-20
Krishna/add graph to self heal

0xKrishna:
# Description

Added sub-graph support in self-heal mechanism, As of now updated 2 methods which uses the limit bases ETH calls.
This introduces a new config flag `sub_graph_url` which if set will enable usage of subgraph for fetching self-heal specific data.

# Changes

# Nodes audience

This will be only used on a couple of internal nodes for self-heal and debugging

# Checklist

- [ ] I have added at least 2 reviewer or the whole pos-v1 team
- [ ] I will be resolving comments - if any - by pushing each fix in a separate commit and linking the commit hash in the comment reply

# Cross repository changes

  - In case link the PR here:
  - In case link the PR here:

## Testing

- [ ] I have added unit tests
- [ ] I have tested this code manually on mumbai

### Manual tests

# Additional comments


bridge/setu/listener/rootchain_selfheal.go
```diff
@@ -199,32 +208,26 @@ func (rl *RootChainListener) processStateSynced(ctx context.Context) {
...
 	}
 
 	if time.Since(blockTime) < helper.GetConfig().SHMaxDepthDuration {
-		rl.Logger.Info("Block time is less than an hour, skipping state sync")
+		rl.Logger.Info("Block time is less than max time depth, skipping event")
 		return true, err
 	}
 
-	topic := event.Topics[0].Bytes()
+	topic := vLog.Topics[0].Bytes()
```
manav2401, line 223: I've seen before that vLog i.e. event instance was empty in some cases (mentioned in one of the comment above). Is it still the case or we can safely skip that check. Just a minor concern because this statement here will panic if it's empty.

manav2401 (approved):

bridge/setu/listener/rootchain_selfheal.go
0xKrishna: It's safe now, I found the reason why it was empty earlier and fixed that.

VAIBHAVJINDAL3012 (approved):

PR #1039, opened 2023-08-24
Combine multiple hardforks into one

VAIBHAVJINDAL3012:
# Description
This PR merges all the following hf
1.GetValidatorSetRotationStopHeight
2.GetMilestoneHardForkHeight()
3.GetNoAckProposerCheckHeight()
into the one single aalborg hf

### DIFF
--- a/checkpoint/side_milestone_handler.go
+++ b/checkpoint/side_milestone_handler.go
@@ -22,12 +22,6 @@ func SideHandleMsgMilestone(ctx sdk.Context, k Keeper, msg types.MsgMilestone, c
 	// logger
 	logger := k.Logger(ctx)
 
-	//Check whether the chain has reached the hard fork length
-	if ctx.BlockHeight() < helper.GetMilestoneHardForkHeight() {
-		logger.Error("Network hasn't reached the", "Hard forked height", helper.GetMilestoneHardForkHeight())
-		return common.ErrorSideTx(k.Codespace(), common.CodeInvalidBlockInput)
-	}
-
 	//Get the milestone count
 	count := k.GetMilestoneCount(ctx)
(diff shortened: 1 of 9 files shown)


checkpoint/side_milestone_handler.go
```diff
@@ -22,12 +22,6 @@ func SideHandleMsgMilestone(ctx sdk.Context, k Keeper, msg types.MsgMilestone, c
 	// logger
 	logger := k.Logger(ctx)
 
-	//Check whether the chain has reached the hard fork length
-	if ctx.BlockHeight() < helper.GetMilestoneHardForkHeight() {
```
temaniarpit27, line 26: why have we removed this?

checkpoint/side_milestone_handler_test.go
```diff
@@ -33,27 +33,6 @@ func (suite *SideHandlerTestSuite) TestSideHandleMsgMilestone() {
 
 	borChainId := "1234"
 
-	suite.Run("Failure-Before Hard Fork", func() {
```
temaniarpit27, line 36: why removed this?

checkpoint/side_milestone_handler.go
VAIBHAVJINDAL3012: This is redundant, Msg will never reach this if the chain has not crossed the hardfork.

checkpoint/side_milestone_handler_test.go
VAIBHAVJINDAL3012: This was related to HF check in sidehandler.

temaniarpit27 (approved):

manav2401 (approved):

PR #1153, opened 2024-03-28
checkpoint: fix numerical key ordering

taratorio:
# Description

Problem: we would like to use the checkpoints/list heimdall provides to fetch checkpoints in bulk. However we noticed one issue with it. The list of checkpoints returned on a page are not ordered correctly.

Example: https://heimdall-api.polygon.technology/checkpoints/list?page=1&limit=10
Screenshot 2024-03-28 at 13 08 18

After digging through the code I suspect the issue is coming from how we persist the checkpoint number key in the DB. Under the hood I believe that the store uses an AVL tree (or some sort of balanced binary tree) and the iterator used in the checkpoint/list API traverses that tree in order based on the key. 

Unfortunately currently we store the key using `[]byte(strconv.FormatUint(checkpointNumber, 10))` which results in a lexicographical sort in the AVL tree instead of a numerical one (i.e. checkpoint number 2 > checkpoint number 10). I suspect we need to use `sdk.Uint64ToBigEndian(checkpointNumber)` instead.

This is a simple fix however it necessitates a data migration - all existing checkpoint key-value pairs need to be moved to a new key-value pair using the correct numerical byte representation of the key. There are only 60k checkpoints on mainnet so such a migration should in theory be quick and unnoticeable.

I'm not sure what the best way to apply such migrations in the cosmos sdk is so please do let me know if you have any suggestions of how to do the migration in a better way than the current solution. Also note, this is pending testing which I haven't started yet - I first would like to start the conversation and get some initial feedback whether this is the right direction before investing more time into this.

# Changes

# Breaking changes

Please complete this section if any breaking changes have been made, otherwise delete it

# Nodes audience

In case this PR includes changes that must be applied only to a subset of nodes, please specify how you handled it (e.g. by adding a flag with a default value...)

# Checklist

- [] I have added at least 2 reviewer or the whole pos-v1 team
- [ ] I will be resolving comments - if any - by pushing each fix in a separate commit and linking the commit hash in the comment reply

# Cross repository changes

  - In case link the PR here:
  - In case link the PR here:

## Testing

- [ ] I have added unit tests

### Manual tests

# Additional comments

### DIFF
--- a/checkpoint/genesis.go
+++ b/checkpoint/genesis.go
@@ -2,7 +2,6 @@ package checkpoint
 
 import (
 	"errors"
-	"fmt"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
@@ -44,10 +43,6 @@ func InitGenesis(ctx sdk.Context, keeper Keeper, data types.GenesisState) {
 		if err := keeper.SetCheckpointBuffer(ctx, *data.BufferedCheckpoint); err != nil {
 			keeper.Logger(ctx).Error("InitGenesis | SetCheckpointBuffer", "error", err)
 		}
-	}
-
-	if err := keeper.MigrateCheckpointKeyV2(ctx); err != nil {
-		panic(fmt.Errorf("checkpoint key v2 migration failed: %w", err))
 	}
 
 	// Set initial ack count
--- a/checkpoint/genesis.go
+++ b/checkpoint/genesis.go
@@ -2,6 +2,7 @@ package checkpoint
 
 import (
 	"errors"
+	"fmt"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
@@ -43,6 +44,10 @@ func InitGenesis(ctx sdk.Context, keeper Keeper, data types.GenesisState) {
 		if err := keeper.SetCheckpointBuffer(ctx, *data.BufferedCheckpoint); err != nil {
 			keeper.Logger(ctx).Error("InitGenesis | SetCheckpointBuffer", "error", err)
 		}
+	}
+
+	if err := keeper.MigrateCheckpointKeyV2(ctx); err != nil {
+		panic(fmt.Errorf("checkpoint key v2 migration failed: %w", err))
 	}
 
 	// Set initial ack count
--- a/checkpoint/keeper.go
+++ b/checkpoint/keeper.go
@@ -19,10 +19,11 @@ import (
 var (
 	DefaultValue = []byte{0x01} // Value to store in CacheCheckpoint and CacheCheckpointACK & ValidatorSetChange Flag
 
-	ACKCountKey         = []byte{0x11} // key to store ACK count
-	BufferCheckpointKey = []byte{0x12} // Key to store checkpoint in buffer
-	CheckpointKey       = []byte{0x13} // prefix key for when storing checkpoint after ACK
-	LastNoACKKey        = []byte{0x14} // key to store last no-ack
+	ACKCountKey             = []byte{0x11} // key to store ACK count
+	BufferCheckpointKey     = []byte{0x12} // Key to store checkpoint in buffer
+	CheckpointKeyDeprecated = []byte{0x13} // use CheckpointKey instead
+	LastNoACKKey            = []byte{0x14} // key to store last no-ack
+	CheckpointKey           = []byte{0x15} // prefix key for when storing checkpoint after ACK
 )
 
 // ModuleCommunicator manages different module interaction
@@ -191,7 +192,7 @@ func (k *Keeper) GetLastCheckpoint(ctx sdk.Context) (hmTypes.Checkpoint, error) 
 
 // GetCheckpointKey appends prefix to checkpointNumber
 func GetCheckpointKey(checkpointNumber uint64) []byte {
-	checkpointNumberBytes := []byte(strconv.FormatUint(checkpointNumber, 10))
+	checkpointNumberBytes := sdk.Uint64ToBigEndian(checkpointNumber)
 	return append(CheckpointKey, checkpointNumberBytes...)
 }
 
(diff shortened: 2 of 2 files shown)


checkpoint/genesis.go
```diff
@@ -45,6 +46,10 @@ func InitGenesis(ctx sdk.Context, keeper Keeper, data types.GenesisState) {
 		}
 	}
 
+	if err := keeper.MigrateCheckpointKeyV2(ctx); err != nil {
```
taratorio, line 49: TODOs:
- is this a good place to run the migration?
- is InitGenesis guaranteed to be run on process restart?
- test and see how long it takes - first impression is that this should be quick to complete given the level db table has only 60k entries in bor-mainnet

checkpoint/keeper.go
```diff
@@ -191,7 +192,7 @@ func (k *Keeper) GetLastCheckpoint(ctx sdk.Context) (hmTypes.Checkpoint, error)
 
 // GetCheckpointKey appends prefix to checkpointNumber
 func GetCheckpointKey(checkpointNumber uint64) []byte {
-	checkpointNumberBytes := []byte(strconv.FormatUint(checkpointNumber, 10))
+	checkpointNumberBytes := sdk.Uint64ToBigEndian(checkpointNumber)
```
taratorio, line 195: this is the actual fix for having correct numerical ordering

checkpoint/genesis.go
Raneet10: `InitGenesis` runs during the genesis of the chain. Hence a restart won't work I'm afraid. 
The only plausible solution here it seems would be to start a new chain with the migrated checkpoint keys.
While heimdall-v2  is in the works (with the consensus engine upgraded to [CometBFT](https://github.com/cometbft/cometbft) and Cosmos SDK to `v0.50 `) and we would need to do a "hard fork" (i.e starting a new chain with migrated data as genesis), that's still gonna take a couple of months.

taratorio: @Raneet10 thank you for the feedback. In that case the alternative I may suggest is we do the migration via the existing APIs. E.g.
- persist new checkpoints using the new sorted key
- whenever a checkpoint by id or checkpoint list is queried via the keeper - we perform the migration then as an one-off operation 
- use the new key for querying

From a user perspective it should be seamless. I believe both Bor and Erigon would query either the checkpoints/<id> and/or checkpoints/list APIs. 

I'll prep the code over the next few days when I have some time and circle back.

taratorio: @Raneet10 - just pushed new approach - what do you think?

Raneet10: > @Raneet10 - just pushed new approach - what do you think?

Thanks @taratorio , I'll have a look soon. 


> @Raneet10 thank you for the feedback. In that case the alternative I may suggest is we do the migration via the existing APIs. E.g.
> 
> * persist new checkpoints using the new sorted key
> * whenever a checkpoint by id or checkpoint list is queried via the keeper - we perform the migration then as an one-off operation
> * use the new key for querying
> 
> From a user perspective it should be seamless. I believe both Bor and Erigon would query either the checkpoints/id and/or checkpoints/list APIs.
> 
> I'll prep the code over the next few days when I have some time and circle back.

Apologies for not mentioning it in my previous comment but the problem here is that all the validators would have to have the keys get rearranged at the same time otherwise we probably risk of seeing a mismatch occurring in the merkle root hash of application (i.e. Heimdall) state tree (`appHash`) which will break the consensus. 
This needs more careful thought, but maybe, for this one time, we introduce a transaction type in `checkpoint` for this purpose and if the tx gets enough votes (and it hopefully should) from validators, we will execute this migration on all the validator machines. WDYT ?

taratorio: @Raneet10 oh I see, that makes sense. I was not aware that there is such a state hash at the application level. Do you have a pointer to documentation I can read about this? I'm happy to do the work and re-adapt the solution if you think it is doable and provide some high level guidance.

Raneet10: Hey @taratorio, sorry for the late reply. If you're looking for a general overview of ABCI (interface via which tendermint interacts with the application, which is heimdall for us) you can maybe start [here](https://github.com/maticnetwork/tendermint/blob/2f132b99d76314ddb1b73d4888da874a3f345ab0/docs/spec/abci/README.md). Although, please note that this has been updated in favour of [ABCI 2.0](https://github.com/cometbft/cometbft/blob/main/spec/abci/abci%2B%2B_basic_concepts.md) now, which we will be switching to soon with heimdall v2. 

If you're looking to add a tx type that results in the migration, as an example you can refer to [MsgCheckpoint](https://github.com/maticnetwork/heimdall/blob/ae5a160feaf278b0127c3e2b45bb1338b1de9d7a/checkpoint/types/msg.go#L95C6-L95C19). I'll try to illustrate the flow here:

-  A validator [initiates and broadcasts](https://github.com/maticnetwork/heimdall/blob/ae5a160feaf278b0127c3e2b45bb1338b1de9d7a/checkpoint/client/cli/tx.go#L144) the `MsgCheckpoint` tx. In practice this is automatically handled by the [bridge process](https://github.com/maticnetwork/heimdall/blob/master/bridge/setu/processor/checkpoint.go). 
- Each validator process runs some basic validation in the [msg handler](https://github.com/maticnetwork/heimdall/blob/ae5a160feaf278b0127c3e2b45bb1338b1de9d7a/checkpoint/handler.go#L79). 
- Each validator then votes on the correctness of the data with a [side handler ](https://github.com/maticnetwork/heimdall/blob/ae5a160feaf278b0127c3e2b45bb1338b1de9d7a/checkpoint/side_handler.go#L95). If it's correct the validator votes  a`YES`. 
- [If we have 2/3+ YES votes](https://github.com/maticnetwork/heimdall/blob/ae5a160feaf278b0127c3e2b45bb1338b1de9d7a/app/side_tx_processor.go#L101-L106), we execute the tx in the [post handler to persist in the app DB](https://github.com/maticnetwork/heimdall/blob/ae5a160feaf278b0127c3e2b45bb1338b1de9d7a/checkpoint/side_handler.go#L277).  

The voting process happens via a mechanism we call sidechannel. You can take a look [here](https://polygontechnology.notion.site/Side-channel-on-Tendermint-628514d015734900a66d32c929026da8) in case you're interested. 
Hope this helps!

taratorio: thanks, as discussed offline we will leave this for the heimdall team to handle as part of heimdall v2

Raneet10:
Closing this PR as we will inculcate these changes into `heimdall-v2`. Thanks again for flagging @taratorio !

PR #1170, opened 2024-06-07
Handle account sequence correctly in bridge

Raneet10:
# Description

The job of the `broadcaster` component is, as the name suggests, to broadcast txs to tendermint. The ante handler is executed during `checkTx` (invoked by tendermint) and enforces some sanity checks on the tx sent (ensuring enough balance, correct signature, etc.) and at successful verification, increments the account sequence (akin to nonce). 
If any of the checks were to fail, the appropriate error code is baked in `sdk.TxResponse`.

Currently, the `broadcaster` doesn't account for a failed sanity clause (and only checks for errors that pop up _during_ the broadcast itself). This results in an incorrect increment of the bridge's "cached" account sequence and hence causes mismatch with the DB. This PR aims to fix this. 

# Changes

# Checklist

## Testing

- [x] I have added unit tests

### DIFF
--- a/bridge/setu/broadcaster/broadcaster.go
+++ b/bridge/setu/broadcaster/broadcaster.go
@@ -58,7 +58,7 @@ func NewTxBroadcaster(cdc *codec.Codec) *TxBroadcaster {
 }
 
 // BroadcastToHeimdall broadcast to heimdall
-func (tb *TxBroadcaster) BroadcastToHeimdall(msg sdk.Msg, event interface{}) error {
+func (tb *TxBroadcaster) BroadcastToHeimdall(msg sdk.Msg, event interface{}, testOpts ...*helper.TestOpts) (sdk.TxResponse, error) {
 	tb.heimdallMutex.Lock()
 	defer tb.heimdallMutex.Unlock()
 	defer util.LogElapsedTimeForStateSyncedEvent(event, "BroadcastToHeimdall", time.Now())
@@ -75,9 +75,9 @@ func (tb *TxBroadcaster) BroadcastToHeimdall(msg sdk.Msg, event interface{}) err
 		WithSequence(tb.lastSeqNo).
 		WithChainID(chainID)
 
-	txResponse, err := helper.BuildAndBroadcastMsgs(tb.CliCtx, txBldr, []sdk.Msg{msg})
-	if err != nil {
-		tb.logger.Error("Error while broadcasting the heimdall transaction", "error", err)
+	txResponse, err := helper.BuildAndBroadcastMsgs(tb.CliCtx, txBldr, []sdk.Msg{msg}, testOpts...)
+	if err != nil || txResponse.Code != uint32(sdk.CodeOK) {
+		tb.logger.Error("Error while broadcasting the heimdall transaction", "error", err, "txResponse", txResponse.Code)
 
 		// current address
 		address := hmTypes.BytesToHeimdallAddress(helper.GetAddress())
@@ -86,13 +86,13 @@ func (tb *TxBroadcaster) BroadcastToHeimdall(msg sdk.Msg, event interface{}) err
 		account, errAcc := util.GetAccount(tb.CliCtx, address)
 		if errAcc != nil {
 			tb.logger.Error("Error fetching account from rest-api", "url", helper.GetHeimdallServerEndpoint(fmt.Sprintf(util.AccountDetailsURL, helper.GetAddress())))
-			return errAcc
+			return txResponse, errAcc
 		}
 
 		// update seqNo for safety
 		tb.lastSeqNo = account.GetSequence()
 
-		return err
+		return txResponse, err
 	}
 
 	txHash := txResponse.TxHash
@@ -102,7 +102,7 @@ func (tb *TxBroadcaster) BroadcastToHeimdall(msg sdk.Msg, event interface{}) err
 	// increment account sequence
 	tb.lastSeqNo += 1
 
-	return nil
+	return txResponse, nil
 }
 
 // BroadcastToMatic broadcast to matic
--- a/bridge/setu/broadcaster/broadcaster_test.go
+++ b/bridge/setu/broadcaster/broadcaster_test.go
@@ -1,61 +1,296 @@
 package broadcaster
 
 import (
-	"os"
-	"strconv"
+	"bytes"
+	"fmt"
+	"io"
+	"math/big"
+	"net/http"
 	"testing"
 
+	cosmosCtx "github.com/cosmos/cosmos-sdk/client/context"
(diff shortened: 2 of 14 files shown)


cffls (approved):
LGTM! Thanks for adding the tests.

bridge/setu/broadcaster/broadcaster_test.go
```diff
@@ -1,61 +1,296 @@
...
+
+			op:         nil,
+			expResCode: 0,
+			expErr:     false,
+		},
+		{
+			name: "failed broadcast (insufficient funds for fees)",
+			msg:  msgs[1],
+			op: func(hApp *app.HeimdallApp) error {
+				acc := hApp.AccountKeeper.GetAccount(sdkCtx, hmTypes.BytesToHeimdallAddress([]byte(address)))
```
marcello33, line 158: minor one: the `[]byte` type conversion (used in many places in this file) is redundant

helper/test_helpers.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	abci "github.com/tendermint/tendermint/abci/types"
+)
+
+type TestOpts struct {
+	app     abci.Application
+	chainId string
+}
+
+func NewTestOpts(app abci.Application, chainId string) *TestOpts {
```
marcello33, line 12: All these functions are only used in `broadcaster_test.go`. How about we move this file under `./broadcaster` and/or call it `broadcaster_util.go`?

helper/test_helpers.go
```diff
@@ -0,0 +1,33 @@
...
+		app:     app,
+		chainId: chainId,
+	}
+}
+
+func (t *TestOpts) SetApplication(app abci.Application) {
+	t.app = app
+}
+
+func (t *TestOpts) GetApplication() abci.Application {
```
marcello33, line 23: `GetApplication` and `SetChainId` are not used, right? Maybe we can remove them

bridge/setu/broadcaster/broadcaster.go
```diff
@@ -75,9 +75,9 @@ func (tb *TxBroadcaster) BroadcastToHeimdall(msg sdk.Msg, event interface{}) err
 		WithSequence(tb.lastSeqNo).
 		WithChainID(chainID)
 
-	txResponse, err := helper.BuildAndBroadcastMsgs(tb.CliCtx, txBldr, []sdk.Msg{msg})
-	if err != nil {
-		tb.logger.Error("Error while broadcasting the heimdall transaction", "error", err)
+	txResponse, err := helper.BuildAndBroadcastMsgs(tb.CliCtx, txBldr, []sdk.Msg{msg}, testOpts...)
+	if err != nil || txResponse.Code != uint32(sdk.CodeOK) {
```
marcello33, line 79: Digging down the `helper.BuildAndBroadcastMsgs(...)` function, it ultimately invokes `NewResponseFormatBroadcastTx`, which - in case of errors - can return and empty `txResponse`, hence `txResponse.Code` could cause a `nil` pointer exception.

bridge/setu/broadcaster/broadcaster_test.go
```diff
@@ -1,61 +1,296 @@
...
+			if tc.tearDown != nil {
+				err := tc.tearDown(heimdallApp)
+				require.NoError(t, err)
+			}
 		})
 	}
 }
+
+func createTestApp(isCheckTx bool, testOpts *helper.TestOpts) (*app.HeimdallApp, sdk.Context, cosmosCtx.CLIContext) {
+	app := app.Setup(isCheckTx, testOpts)
```
marcello33, line 216: maybe we want to change this var name, as it collides with the imported package name

helper/util.go
```diff
@@ -344,10 +358,18 @@ func BuildAndBroadcastMsgsWithCLI(cliCtx context.CLIContext, txBldr authTypes.Tx
...
+	msgs []sdk.Msg,
+	testOpts ...*TestOpts,
+) ([]byte, error) {
 	// just simulate (useful for testing)
 	if cliCtx.Simulate {
-		return nil, nil
+		if len(testOpts) == 0 || testOpts[0].chainId == "" {
+			return nil, nil
+		}
+		txBldr := txBldr.WithChainID(testOpts[0].chainId)
```
marcello33, line 371: I don't think you need the assignment `:` here

bridge/setu/broadcaster/broadcaster.go
Raneet10: The `err != nil` would account for that no ?

helper/test_helpers.go
Raneet10: I tried that but it ends up causing an import cycle since this code will be imported in the `helper` package (which is itself being imported in `broadcaster`)

helper/test_helpers.go
Raneet10: We can but it looks complete to have both sets of getters and setters. Besides, it might be used later in other tests.

bridge/setu/broadcaster/broadcaster.go
marcello33: If `err != nil` won't be evaluated, correct.
However, at line 80, you try to log `txResponse.Code` when `txResponse` is `nil`
`tb.logger.Error("Error while broadcasting the heimdall transaction", "error", err, "txResponse", txResponse.Code)`

Raneet10: Oh and btw returning an empty `TxResponse` would just mean all the fields are set to default value, in case of  `txResponse.Code` it'd be 0. I don't think it'll cause a nil pointer exception.

helper/util.go
Raneet10: Removed [89c97b2](https://github.com/maticnetwork/heimdall/pull/1170/commits/89c97b2b0a4f76edcdba870ffa0e9521614c90cf).

bridge/setu/broadcaster/broadcaster_test.go
Raneet10: Deduped: [89c97b2](https://github.com/maticnetwork/heimdall/pull/1170/commits/89c97b2b0a4f76edcdba870ffa0e9521614c90cf)

bridge/setu/broadcaster/broadcaster.go
marcello33: Correct. imho it'd still be inconsistent to print `txResponse, 0` but as long as there's no risk of failure, up to you on how to proceed. Thanks

Raneet10: Yeah looks harmless but please do suggest better logging if possible here.

marcello33 (approved):

PR #1195, opened 2024-10-16
Optimization of export genesis command

avalkov:
Optimization of export genesis command to use less memory.


pratikspatil024:
QQ - this is only optimising the clerk event records, right? The other modules does not require any optimisation?

avalkov:
> QQ - this is only optimising the clerk event records, right? The other modules does not require any optimisation?

clerk event records hold like 90% of the entire state so the streamed reading of data is implemented only there.

The second big optimization is how we write the file to disk.

pratikspatil024 (approved):

PR #1201, opened 2024-10-29
Enhance GetBridgeDBInstance: Add Error Handling for LevelDB Initialization

vinayak0035:
#### Overview

Currently, the `GetBridgeDBInstance` function does not handle scenarios where `bridgeDB` may be nil. If the database fails to open for any reason, returning a nil object can lead to potential runtime panics or errors when the returned database instance is used. Adding error handling will help prevent these issues and improve the robustness of the code.

#### Changes Made
- Added error handling within the `sync.Once` block to capture and return any errors encountered while opening the LevelDB database.
  
#### Code Changes
Before:
```go
func GetBridgeDBInstance(filePath string) *leveldb.DB {
    bridgeDBOnce.Do(func() {
        bridgeDB, _ = leveldb.OpenFile(filePath, nil)
    })
    return bridgeDB
}
```

After:
```go
func GetBridgeDBInstance(filePath string) *leveldb.DB {
	bridgeDBOnce.Do(func() {
		var err error
		bridgeDB, err = leveldb.OpenFile(filePath, nil)
		if err != nil {
			log.Fatalln("Error in Bor Opening Database", err.Error())
		}
	})
	return bridgeDB
}
```

#### Benefits
- Enhanced reliability of the application.
- Easier debugging by providing clear error messages.
- Improved user experience by avoiding crashes due to nil references.

### DIFF
--- a/bridge/setu/util/db.go
+++ b/bridge/setu/util/db.go
@@ -1,6 +1,7 @@
 package util
 
 import (
+	"log"
 	"sync"
 
 	"github.com/syndtr/goleveldb/leveldb"
@@ -13,9 +14,12 @@ var bridgeDBCloseOnce sync.Once
 // GetBridgeDBInstance get sington object for bridge-db
 func GetBridgeDBInstance(filePath string) *leveldb.DB {
 	bridgeDBOnce.Do(func() {
-		bridgeDB, _ = leveldb.OpenFile(filePath, nil)
+		var err error
+		bridgeDB, err = leveldb.OpenFile(filePath, nil)
+		if err != nil {
+			log.Fatalln("Error in Opening Database", err.Error())
+		}
 	})
-
 	return bridgeDB
 }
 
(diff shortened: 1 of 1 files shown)


pratikspatil024 (approved):

marcello33:
Hey @vinayak0035, thanks for you contribution.
In your PR's description I see 
`Updated the GetBridgeDBInstance function signature to return an error alongside the database instance.`
But the code does not reflect that. As it is currently, you're only logging an error in case levelDB is not able to open the file, but even with such change `bridgeDB` might be `nil` (hence the crash/panic would still occur).
Is that intended? Thank you.

avalkov:
I am ok with the fix, @vinayak0035 can you just update your description as your not updating the signature (which I agree is redundant because we open the db just once and in all other cases we would just return nil).

vinayak0035:
Hey @avalkov I updated the description , removed the signature.

marcello33 (approved):

PR #1202, opened 2024-10-31
Limit bytes during response body read / NPE fix

marcello33:
# Description

In this PR we provide a bytes limit for the heimdall API responses.
Also, we fix a NPE during smart contracts' calls.

# Changes

# Breaking changes

Please complete this section if any breaking changes have been made, otherwise delete it

# Nodes audience

In case this PR includes changes that must be applied only to a subset of nodes, please specify how you handled it (e.g. by adding a flag with a default value...)

# Checklist

# Cross repository changes

  - In case link the PR here:
  - In case link the PR here:

## Testing

- [ ] I have added unit tests
- [x] I have tested this code manually on local environment

### Manual tests

# Additional comments

### DIFF
--- a/bridge/setu/listener/rootchain_selfheal_graph.go
+++ b/bridge/setu/listener/rootchain_selfheal_graph.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/maticnetwork/heimdall/helper"
 	"io"
 	"math/big"
 	"net/http"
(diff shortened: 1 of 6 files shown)


chetanyb (approved):

cffls (approved):

GumnaamSaaya (commented):
So, the issue before was that in the `GetValidatorInfo` function, `stakerDetails.Status` was being accessed without checking if `stakerDetails` was `nil`. This obviously caused a `nil` pointer dereference and could crash the app if  `GetStakerDetails` returned an error. Basically, big panic vibes
In the updated code, this issue has been fixed they removed `stakerDetails.Status` from the error handling block, and now it just logs the error directly, which is the right move. The `nil` check situation is all sorted now.
also, they improved the error handling in the `GetSpanDetails` function. If there's an error, it’ll properly return `nil` instead of causing a crash. Another thing I noticed: they renamed `GetRootHash` to `GetVoteOnHash`, which honestly makes more sense and makes the code easier to understand.
Overall, this looks like a solid fix the original issue is resolved ;)))

PR #1248, opened 2025-02-21
Fix seed config typos

avalkov:
# Description

Fix typos in seed peers config.


pratikspatil024 (commented):
Nice catch!

Can you also update the mainnet `config.toml`.
I copied these from that file.

Also, I guess this same issue will be present in v2 as well. Good to update those values as well.

pratikspatil024 (approved):

PR #1262, opened 2025-03-21
V1.2.1 candidate

marcello33:
# Description

Candidate branch for `v1.2.1`

# Changes

# Checklist

## Testing

- [ ] I have added unit tests


bor/selection.go
```diff
@@ -65,13 +72,21 @@ func SelectNextProducers(blkHash common.Hash, spanEligibleValidators []hmTypes.V
...
+		return nil, fmt.Errorf("seed value out of range for int64: %d", seedUint64)
+	}
+	seed := int64(seedUint64)
 	// nolint: staticcheck
 	rand.Seed(seed)
 
 	// weighted range from validators' voting power
 	votingPower := make([]uint64, len(spanEligibleValidators))
 	for idx, validator := range spanEligibleValidators {
+		if validator.VotingPower < 0 {
```
avalkov, line 87: Are we sure that it is safer to return error than not? I see we dont have such case but still maybe just skip the -1 and log error?

types/validator.go
```diff
@@ -193,7 +200,10 @@ func (valID ValidatorID) Bytes() []byte {
 
 // Int converts validator ID to int
 func (valID ValidatorID) Int() int {
```
avalkov, line 202: Isnt something like this cleaner?
`if valID > ValidatorID(math.MaxInt) {`

avalkov (approved):

avalkov (changes requested):
Please review my comments.

marcello33:
Postponing

PR #1278, opened 2025-05-06
Bor without heimdall

avalkov:
# Description

Edge cases for span proposition when bor is running without heimdall.


marcello33 (approved):

avalkov:
> So we are trying not to propose spans close to halt height, right?

Yes

